PR #38, opened 2023-03-03
Issue-343 : add configurable prefix to kafka topics

sahibamittal:
### Description

Use custom prefixes (optional) for Kafka topics.

### Addressed Issue

Issue: https://github.com/DependencyTrack/hyades/issues/343

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/event/kafka/KafkaTopic.java
+++ b/src/main/java/org/dependencytrack/event/kafka/KafkaTopic.java
@@ -1,5 +1,8 @@
 package org.dependencytrack.event.kafka;
 
+import io.smallrye.config.SmallRyeConfig;
+import org.eclipse.microprofile.config.ConfigProvider;
+
 public enum KafkaTopic {
 
     NOTIFICATION_ANALYZER("dtrack.notification.analyzer"),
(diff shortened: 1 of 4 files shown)


src/main/java/org/dependencytrack/event/kafka/KafkaTopic.java
```diff
@@ -33,7 +36,12 @@ public enum KafkaTopic {
     }
 
     public String getName() {
-        return name;
+        var prefix = "";
+        SmallRyeConfig config = ConfigProvider.getConfig().unwrap(SmallRyeConfig.class);
+        var prefixConfig = config.getConfigValue("api.topic.prefix");
+        if (prefixConfig.getValue() != null)
+            prefix = prefixConfig.getValue();
+        return prefix + name;
```
nscuro, line 44: There already is a mechanism to read configuration, no need to introduce SmallRye Config as additional dependency. See [`ConfigKey`](https://github.com/DependencyTrack/hyades-apiserver/blob/main/src/main/java/org/dependencytrack/common/ConfigKey.java), which can be accessed with `Config.getInstance().getProperty(ConfigKey.FOOBAR)`.

nscuro (changes requested):

nscuro (approved):

PR #56, opened 2023-03-15
Add missing check for existing vulnerabilities in component

sahibamittal:
### Description

Add missing check for existing vulnerabilities in component to keep audit vulnerabilities.

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
+++ b/src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
@@ -151,10 +151,12 @@ public class VulnerabilityScanResultProcessor extends ContextualFixedKeyProcesso
             query.setFilter("uuid == :uuid");
             query.setParameters(componentUuid);
             Component component = query.executeUnique();
-            component.addVulnerability(vuln);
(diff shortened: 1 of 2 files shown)


mehab (dismissed):

src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
```diff
@@ -151,10 +151,12 @@ private void addVulnerability(final PersistenceManager pm, final UUID componentU
...
-            component = pm.makePersistent(component);
-            var finding = new FindingAttribution(component, vuln, getAnalyzerIdentity(scanner), null, null);
-            pm.makePersistent(finding);
+            if(!contains(vuln, component)) {
+                component.addVulnerability(vuln);
+                component = pm.makePersistent(component);
+                var finding = new FindingAttribution(component, vuln, getAnalyzerIdentity(scanner), null, null);
+                pm.makePersistent(finding);
+            }
             trx.commit();
```
nscuro, line 160: The `commit` can also be moved into the `if` block. If we don't modify anything, there's no point in committing the transaction. It will be rolled back in the `finally` block.

nscuro (approved):

PR #105, opened 2023-04-18
add keystore mapping

sahibamittal:
### Description

Add optional keystore mapping for SSL authentication.

### Checklist


nscuro:
Miscellaneous question: Should we rename the `ssl` properties to `tls`?

In the [Kafka docs](https://docs.confluent.io/platform/current/kafka/authentication_ssl.html#tls-overview) they say they're keeping `ssl` for historical reasons, but they use `ssl` over `tls` consistently, and don't mix them.

> Secure Sockets Layer (SSL) was the predecessor of Transport Layer Security (TLS), and has been deprecated since June 2015. For historical reasons, SSL is used in configuration and code instead of TLS.

PR #126, opened 2023-04-26
Better exception logging and stack propagation

sahibamittal:
### Description

Enhancement includes 

- Ability to provide CacheStampedeBlocker with a list of exceptions to ignore.
- Better exception logging with appropriate level and stack propagation.

### Addressed Issue

https://github.com/DependencyTrack/hyades/issues/374

### Checklist


nscuro:
I think we can even delete this class entirely, as its original purpose was to help with repo meta analysis caching. We have solved this problem differently now. @sahibamittal Are there any more references to this class, or is it safe to delete?

sahibamittal:
> I think we can even delete this class entirely, as its original purpose was to help with repo meta analysis caching. We have solved this problem differently now. @sahibamittal Are there any more references to this class, or is it safe to delete?

Right, this class is not being used anywhere, just in test class. I'll remove it.

nscuro (approved):

PR #146, opened 2023-05-02
Disable DataNucleus L2 cache for BOM upload processing

nscuro:
### Description

Having L2 cache enabled here means that every single component modified by the task will be cached. This can cause excessive heap utilization and increased pressure on the GC when many BOMs are uploaded in close-ish succession.

L2 keeps soft references to cached objects, and it takes a while for those references to decay naturally.

### Addressed Issue

N/A

### Additional Details

N/A

### Checklist


mehab (approved):
lgtm

PR #218, opened 2023-07-04
Improve BOM processing performance

nscuro:
### Description

This refactors `BomUploadProcessingTask` to improve its efficiency and consistency.

Notable changes:

* Components and services are now de-duplicated in-memory based on their `ComponentIdentity` *before* interacting with the database
   * Parsing and converting of the BOM is more performant because it doesn't require interaction with external services (e.g. database)
   * De-duplication takes BOM refs used for the [dependency graph](https://cyclonedx.org/use-cases/#dependency-graph) into consideration. If a component or service "falls victim" to de-duplication, its BOM ref is re-wired to point to the component it was deemed to duplicate instead. This keeps the dependency graph intact.
   * Addresses various issues in the original logic, e.g.
      * https://github.com/DependencyTrack/dependency-track/issues/1905 (regression test included in this PR)
      * https://github.com/DependencyTrack/dependency-track/issues/2131
      * https://github.com/DependencyTrack/dependency-track/issues/2519 (regression test included in this PR)
      * https://github.com/DependencyTrack/dependency-track/issues/2265
* When correlating components from the uploaded BOM with components in the database, the correlation query is now more strict
   * See https://github.com/DependencyTrack/dependency-track/issues/2519#issuecomment-1621889555 for more context
* Processing now runs in a single database transaction
   * Contrasting previous logic which used *many* small transactions
   * Now, if BOM processing fails half-way, we don't leave broken state behind
   * Changes are regularly `flush`ed to the database over the course of the transaction

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/635

### Additional Details

But does it perform better? Uh, yeah.

I compared this PR with the current `snapshot` build of hyades-apiserver, using the [Hyades `docker-compose.yml`](https://github.com/DependencyTrack/hyades/blob/main/docker-compose.yml). Other Hyades services besides the API server have been turned off for this test.

I uploaded the same 1000 BOMs to each version. The high-level results are as follows:

| API Server version | BOM Upload Start | BOM Upload End | BOM Processing End |
| :---- | :----- | :----- | :------ |
| This PR | 13:55:00 | 13:59:06 | 13:59:07 |
| Current `snapshot` | 14:07:45 | 14:18:44 | 14:22:23 |

The API server version built from this PR took just about 4min for the entire operation, whereas the current `snapshot` took 15min.

#### System / JVM Statistics

Overall resource footprint is down. Less CPU and Heap usage. 

As the Postgres container is running on the same machine, the *System* CPU usage graph being lower for this PR is a good sign as well. Less stress is being put on the database server.

##### This PR

Screenshot from 2023-07-06 14-00-15

##### `snapshot`

Screenshot from 2023-07-06 14-24-12

#### Database / ORM Statistics

Because all operations of BOM processing now run in a single transaction, the total number of transaction is noticeably down. Also, only the transactional connection pool is utilized. It can be observed that only at the very beginning there are a few connections in *pending* state.

In contrast, in the `snapshot` version, both connection pools are heavily utilized, and there's a constantly high number of *pending* connections on the non-transactional pool. The number of transactions is also insanely high.

##### This PR

Screenshot from 2023-07-06 14-00-37

##### `snapshot`

Screenshot from 2023-07-06 14-24-45

#### BOM Processing Duration

The average duration of BOM processing stays constantly in the single-digit second range for this PR, whereas for `snapshot` it even reaches up to almost 2min.

##### This PR

Screenshot from 2023-07-06 14-03-08

##### `snapshot`

image

#### Kafka Records Produced

Another nice indicator that things have improved: The number of records produced to Kafka per second have noticeably increased. This is just a side effect of BOM processing completing faster, but still nice to observe.

##### This PR

Screenshot from 2023-07-06 14-01-05

##### `snapshot`

Screenshot from 2023-07-06 14-25-26

#### Internal Event System

The laptop I am testing on has 16 CPU cores. DT will thus use a worker thread pool of up to 64 threads. Worker pool threads are responsible for processing BOMs, thus their usage is a good indicator of how performant the processing is.

With this PR, the number of workers used just barely reaches 20 shorty. Also, there were no *Events Queued* the entire time, indicating that BOMs were processed almost as soon as they were uploaded.

 With `snapshot`, the pool was entirely saturated the whole time. And still events were queueing up, eventually reaching 250 yet-to-be-processed events.

##### This PR

Screenshot from 2023-07-06 14-01-21

##### `snapshot`

Screenshot from 2023-07-06 14-25-41

### Checklist


nscuro:
Note: This PR should be squash-merged so it's easier to revert if we have to.

VithikaS (approved):
It looks great!

nscuro:
@VithikaS Could you please re-approve, given everything still looks OK? Had to resolve some merge conflicts by rebasing.

VithikaS (dismissed):

src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
```diff
@@ -420,19 +420,54 @@ public void recursivelyDelete(Component component, boolean commitIndex) {
...
+        } else {
+            coordinatesFilter += " && version == null";
+        }
+        coordinatesFilter += ")";
+        filterParts.add(coordinatesFilter);
+
+        final var filter = "project == :project && (" + String.join(" && ", filterParts) + ")";
+        params.put("project", project);
+
+        final Query<Component> query = pm.newQuery(Component.class, filter);
```
mehab, line 468: should try with resources be used here?

src/main/java/org/dependencytrack/tasks/BomUploadProcessingTask.java
```diff
@@ -84,174 +114,715 @@ public BomUploadProcessingTask() {
...
-    private void processService(final QueryManager qm, final Bom bom, ServiceComponent service,
-                                final List<ServiceComponent> flattenedServices) {
-        service = qm.createServiceComponent(service, false);
-        final long oid = service.getId();
-        // Refreshing the object by querying for it again is preventative
-        flattenedServices.add(qm.getObjectById(ServiceComponent.class, oid));
-        if (service.getChildren() != null) {
-            for (final ServiceComponent child : service.getChildren()) {
-                processService(qm, bom, child, flattenedServices);
+    private static Map<ComponentIdentity, Component> processComponents(final Context ctx, final QueryManager qm,
```
mehab, line 363: ctx not used in the method here

src/main/java/org/dependencytrack/tasks/BomUploadProcessingTask.java
```diff
@@ -84,174 +114,715 @@ public BomUploadProcessingTask() {
...
+
+        final var vulnAnalysisEvents = new ArrayList<ComponentVulnerabilityAnalysisEvent>();
+        final var repoMetaAnalysisEvents = new ArrayList<ComponentRepositoryMetaAnalysisEvent>();
+
+        try (final var qm = new QueryManager()) {
+            final PersistenceManager pm = qm.getPersistenceManager();
+
+            // Disable reachability checks on commit.
+            // See https://www.datanucleus.org/products/accessplatform_4_1/jdo/performance_tuning.html
+            //
```
mehab, line 213: amazing documentation 💯 thanks for this level of detail <3

src/main/java/org/dependencytrack/tasks/BomUploadProcessingTask.java
```diff
@@ -84,174 +114,715 @@ public BomUploadProcessingTask() {
...
-        if (isNew) {
-            newComponents.add(qm.detach(Component.class, component.getId()));
+    private void processBom(final Context ctx, final File bomFile) throws BomProcessingException {
+        LOGGER.info("Consuming uploaded BOM (%s)".formatted(ctx));
+        final org.cyclonedx.model.Bom cdxBom = parseBom(ctx, bomFile);
+
+        // Keep track of which BOM ref points to which component identity.
+        // During component and service de-duplication, we'll potentially drop
+        // some BOM refs, which can break the dependency graph.
+        final var identitiesByBomRef = new HashMap<String, ComponentIdentity>();
```
mehab, line 166: do we need this to be thread safe?

nscuro: Nope, the event system will create a new task instance for every event, so task instances are never accessed by multiple threads. 

The`Map` is also [local](https://www.baeldung.com/java-local-variables-thread-safe), and not a class-level variable. So even if multiple threads were sharing a task instance, each invocation of `processBom` would have its own `identitiesByBomRef`.

src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
nscuro: Yes, good catch! 

FYI I'm using `try { ... } finally { query.closeAll() }` instead of try-with-resources, because the `close` method of `Query` has a `throws Exception` declaration that "poisons" the code calling it. `closeAll` achieves the same, but doesn't force us to do `catch (Exception)`.

PR #220, opened 2023-07-05
excluded not required transitive dependency

mehab:
### Description

Excluded snakeyaml dependency that is not used by project and still leading to vulnerability issues

### Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -188,6 +188,13 @@
             <groupId>us.springett</groupId>
             <artifactId>alpine-server</artifactId>
             <version>${lib.alpine.version}</version>
+            <exclusions>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -188,6 +188,13 @@
             <groupId>us.springett</groupId>
             <artifactId>alpine-server</artifactId>
             <version>${lib.alpine.version}</version>
+            <exclusions>
+                <!-- this exclusion done to get rid of vulnerability in the project as this transitive dependency is not being used in the project -->
```
nscuro, line 192: Could you include *which* vulnerabilities this is supposed to fix?

mehab: done

PR #222, opened 2023-07-06
Lock singleton tasks

VithikaS:
### Description

This PR locks all singleton tasks so we have only one pod executing tasks

Screenshot 2023-07-05 at 19 37 42

### Addressed Issue

It addresses some of the issues in https://github.com/DependencyTrack/hyades/issues/375

### Checklist


src/main/java/org/dependencytrack/tasks/EpssMirrorTask.java
```diff
@@ -80,11 +90,25 @@ public EpssMirrorTask() {
      */
     public void inform(final Event e) {
         if (e instanceof EpssMirrorEvent && this.isEnabled) {
+            JdbcLockProvider instance = getJdbcLockProviderInstance();
+            LockingTaskExecutor executor = getLockingTaskExecutorInstance(instance);
+            LockConfiguration lockConfiguration = new LockConfiguration(Instant.now(),
+                    EPSS_MIRROR_TASK_LOCK.name(),
+                    Duration.ofMillis(Config.getInstance().getPropertyAsInt(TASK_MIRROR_EPSS_LOCK_AT_MOST_FOR)),
+                    Duration.ofMillis(Config.getInstance().getPropertyAsInt(TASK_MIRROR_EPSS_LOCK_AT_LEAST_FOR)));
+
```
nscuro, line 99: The procedure of:

1. Acquiring the provider instance
2. Creating an executor
3. Creating a lock configuration
4. Running a task using the executor

Is the same everywhere so far. Would it make sense to abstract this in a common `LockProvider` method?

Perhaps even to something like:

```java
LockProvider.executeWithLock(EPSS_MIRROR_TASK_LOCK, () -> ...);
```

The common utility method could perform the mapping of `LockName` to the corresponding `ConfigKey`s.

Going one step further, it could even perform the generic handling of checked exceptions (catching `Exception` and re-throwing as `RuntimeException`).

src/main/java/org/dependencytrack/event/LdapSyncEvent.java
nscuro, line 1: Can't we reuse the existing `LdapSyncEvent` from Alpine?

src/main/java/org/dependencytrack/tasks/InternalComponentIdentificationTask.java
```diff
@@ -72,6 +97,12 @@ private void analyze() throws Exception {
 
             List<Component> components = fetchNextComponentsPage(pm, null);
             while (!components.isEmpty()) {
+                //Extend the lock by 5 min everytime we have a page.
+                //We will get max 500 components in a page
+                //Reason of not extending at the end of loop is if it does not have to do much,
+                //It might finish execution before lock could be extended resulting in error
+                LOGGER.debug("extending lock of internal component identification by 5 min");
+                LockExtender.extendActiveLock(Duration.ofMillis(300000), ZERO);
```
nscuro, line 105: The comment above mentions minutes as time unit, I think this would make it more readable:

```suggestion
                LockExtender.extendActiveLock(Duration.ofMinutes(5), ZERO);
```

nscuro (commented):
Looks good, just a few minor suggestions! 🤏

src/main/java/org/dependencytrack/event/LdapSyncEvent.java
VithikaS: I should actually, I should have just changed the wrapper and used the same event. Thanks, will update it :)

src/main/java/org/dependencytrack/tasks/EpssMirrorTask.java
VithikaS: Hi @nscuro , can you please have a look again? I have not put generic exception handling, not all tasks will have checked exceptions. If it does not look good, i will make change

nscuro: SchedLock also provides a `executeWithLock` [method](https://github.com/lukas-krecan/ShedLock/blob/4958bb83428d06d33d3347e01af271222ca0fc7c/shedlock-core/src/main/java/net/javacrumbs/shedlock/core/LockingTaskExecutor.java#L26) that takes a `LockingTaskExecutor.Task` instead of a `Runnable`, and has a `throws Throwable` declaration.

We could do the same (provide one method with `Runnable` and without `throws Throwable`, and one with `Task` and with `throws Throwable`). 

That way we can run methods without checked exceptions in the `Runnable` variant, and those *with* checked exception in the `Task` variant.

nscuro (approved):

PR #223, opened 2023-07-06
added changes for authentication

mehab:
### Description

Added global authentication for repositories.
Currently a repository only allows for authentication if the packages we need are internal.
However, there might be external packages as well that could only be downloaded from a repo that is authenticated.
With this change, a repository can be marked as internal. At the same time there is a choice of whether it is authenticated or not. If authenticated then user can enter the corresponding credentials for the same

### Checklist


sahibamittal (dismissed):
lgtm

sahibamittal (approved):
LGTM

PR #226, opened 2023-07-07
Add common `hostname` tag to Prometheus metrics

nscuro:
### Description

This PR adds a `hostname` tag to all Prometheus metrics.

For k8s deployments specifically, this can be helpful when multiple instances of the API server are deployed, and differentiating between them is desired.

### Addressed Issue

N/A

### Additional Details

N/A

### Checklist

-~This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/dependency-track/tree/master/docs/_docs) accordingly~

### DIFF
--- a/src/main/java/org/dependencytrack/health/HealthCheckInitializer.java
+++ b/src/main/java/org/dependencytrack/health/HealthCheckInitializer.java
@@ -24,9 +24,14 @@ import alpine.server.health.HealthCheckRegistry;
 import alpine.server.health.checks.DatabaseHealthCheck;
 import io.github.mweirauch.micrometer.jvm.extras.ProcessMemoryMetrics;
 import io.github.mweirauch.micrometer.jvm.extras.ProcessThreadMetrics;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.config.MeterFilter;
 
 import javax.servlet.ServletContextEvent;
 import javax.servlet.ServletContextListener;
+import java.net.InetAddress;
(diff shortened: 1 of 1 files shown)


nscuro:
Label is not consistently applied to all metrics, needs more work.

nscuro:
Hmmm... `MeterFilter`s are only applied to all `Meter`s created *after* the filter was added.

Because Alpine and DT do a lot of initialization in `static` fields, lots of `Meter`s are created at class loading time. If I move the `MeterFilter` registration to the [*first* `ServletContextListener`](https://github.com/DependencyTrack/hyades-apiserver/blob/cd1ae7b1c014966b646c9d83d58131ef5ffeba56/src/main/webapp/WEB-INF/web.xml#L27), it affects more metrics than before, but e.g. the [`EventService` executor service metrics](https://github.com/stevespringett/Alpine/blob/7615930ec767370cc6bd63a7f25f4456d8874752/alpine-infra/src/main/java/alpine/event/framework/EventService.java#L56) are still unaffected.

The registration of `MeterFilter`s likely needs to be done in Alpine, when the `MeterRegistry` is created: https://github.com/stevespringett/Alpine/blob/7615930ec767370cc6bd63a7f25f4456d8874752/alpine-common/src/main/java/alpine/common/metrics/Metrics.java#L33

nscuro:
Implemented in Alpine (https://github.com/stevespringett/Alpine/pull/501), blocked until next Alpine version is released.

mehab (approved):

PR #234, opened 2023-07-11
Substituting fixed delays with cron expression

VithikaS:
### Description

Changed scheduling of tasks from using fixed delays to using cron expression.
This change will enable triggering of jobs on more calendar like events.

### Addressed Issue

Part of work required to make api server highly available 
https://github.com/DependencyTrack/hyades/issues/375

### Checklist


src/main/java/org/dependencytrack/tasks/BaseTaskScheduler.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private static final Logger LOGGER = Logger.getLogger(BaseTaskScheduler.class);
+
+    private Timer timer;
+
+    private static final long POLLING_INTERVAL_IN_MILLIS = 10000;
+
+    protected void scheduleTask(Map<Event, Schedule> eventScheduleMap) {
+        timer = new Timer();
+        timer.schedule(new ScheduleEvent().eventScheduleMap(eventScheduleMap), 100, 10000);
```
nscuro, line 24: Should we make the delay configurable, and use a higher default value than `100`?

Generally I think we should avoid scheduling any tasks while the API server is starting up, but the duration of the startup will of course vary depending on available resources etc.

src/main/java/org/dependencytrack/tasks/TaskScheduler.java
```diff
@@ -41,118 +43,105 @@
...
+                            Map.Entry::getValue));
 
-            // Creates a new event that executes every 6 hours (21600000) by default after an initial 1 hour (3600000) delay
-            scheduleEvent(new InternalComponentIdentificationEvent(), 3600000, getCadenceConfigPropertyValueInMilliseconds(qm, TASK_SCHEDULER_INTERNAL_COMPONENT_IDENTIFICATION_CADENCE));
+            scheduleTask(mergedEventScheduleMap);
 
-            scheduleEvent(new VulnerabilityScanCleanupEvent(), Duration.ofSeconds(30).toMillis(), Duration.ofDays(6).toMillis());
+        } catch (InvalidExpressionException invalidExpressionException) {
+            LOGGER.error("Exception in parsing cron expression and creating schedule", invalidExpressionException);
         }
```
nscuro, line 120: I'd say we should just propagate the exception to the caller, or re-throw it as `RuntimeException`. Ideally, if a cron expression is invalid, we'd want the API server to not even start.

src/main/resources/application.properties
```diff
@@ -365,4 +365,20 @@ task.mirror.epss.lockAtLeastForInMillis=3000
...
+task.cron.componentIdentification=* 0/6 * * *
+task.cron.ldapSync=* 0/6 * * *
+task.cron.vulndbSync=* 0/4 * * *
+task.cron.repoMetaAnalysis=* 0/8 * * *
+task.cron.vulnAnalysis=* 0/8 * * *
+task.cron.vulnScanCleanUp=* 0/8 * * *
+task.cron.fortify.ssc.sync=* 0/8 * * *
+task.cron.defectdojo.sync=* 0/8 * * *
+task.cron.kenna.sync=* 0/8 * * *
+task.cron.index.consistency.check=* 0/8 * * *
```
nscuro, line 384: For the default values we should add comments to explain what they are, as cron expressions can be quite cryptic if you're not used to them. For example, for `task.cron.metrics.portfolio`, add a comment saying `Every 4 hours`.

nscuro (commented):
Thanks @VithikaS, really like the simplicity of this! Looks good so far.

src/main/resources/application.properties
VithikaS: All good comments.. was going to address all of them in next commit. Thanks for the review!

mehab (approved):
lgtm

PR #239, opened 2023-07-17
Implement Kafka Streams exception handlers

nscuro:
### Description

This PR adds custom handlers for deserialization, processing, and production exceptions in Kafka Streams.

### Addressed Issue

Partially addresses https://github.com/DependencyTrack/hyades/issues/674

### Additional Details

See https://github.com/DependencyTrack/hyades/issues/674 for details.

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/event/kafka/exception/KafkaStreamsUncaughtExceptionHandler.java
+++ b/src/main/java/org/dependencytrack/event/kafka/exception/KafkaStreamsUncaughtExceptionHandler.java
@@ -1,0 +1,105 @@
+package org.dependencytrack.event.kafka.exception;
+
+import alpine.Config;
+import alpine.common.logging.Logger;
+import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.apache.http.conn.ConnectTimeoutException;
+import org.apache.kafka.streams.errors.StreamsUncaughtExceptionHandler;
+import org.datanucleus.api.jdo.exceptions.ConnectionInUseException;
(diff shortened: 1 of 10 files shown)


src/main/java/org/dependencytrack/event/kafka/exception/KafkaStreamsUncaughtExceptionHandler.java
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public StreamThreadExceptionResponse handle(final Throwable exception) {
+        final Throwable rootCause = ExceptionUtils.getRootCause(exception);
+
+        if (rootCause instanceof TimeoutException
+                || rootCause instanceof ConnectTimeoutException
```
VithikaS, line 57: Should we add ReadTimeoutException also ?

VithikaS (dismissed):
Looks good, just added a minor question in PR.

src/main/java/org/dependencytrack/event/kafka/exception/KafkaStreamsUncaughtExceptionHandler.java
nscuro: Good point, will add it!

nscuro: Hmmm, there exists no `ReadTimeoutException`... But I do also recall seeing this class before, but perhaps in another code base.

VithikaS (approved):

PR #241, opened 2023-07-18
Create model for workflow engine

VithikaS:
### Description

It creates model for workflow state according to design considerations https://github.com/DependencyTrack/hyades/issues/664

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/671 

### Checklist


src/main/java/org/dependencytrack/model/WorkflowState.java
```diff
@@ -0,0 +1,136 @@
...
+
+    @Persistent
+    @Column(name = "STEP", jdbcType = "VARCHAR", length = 64, allowsNull = "false")
+    @NotNull
+    private String step;
+
+    @Persistent
+    @Column(name = "STATUS", jdbcType = "VARCHAR", length = 64, allowsNull = "false")
+    @NotNull
+    private String status;
```
nscuro, line 67: I think it'd make sense to use enums for these fields. 

For `status` in particular, this would make it possible to encode the state machine behavior, e.g. which state is terminal, whether it is possible to transition from state A to state B etc.

src/main/java/org/dependencytrack/model/WorkflowState.java
```diff
@@ -0,0 +1,136 @@
...
+
+    public enum FetchGroup {
+        ALL
+    }
+
+    @PrimaryKey
+    @Persistent(valueStrategy = IdGeneratorStrategy.NATIVE)
+    private long id;
+
+    // null allowed because first step will have no parent and will be the first step of recursion
```
nscuro, line 40: True, marking this an non-nullable in the schema draft in https://github.com/DependencyTrack/hyades/issues/664 was a mistake. Just corrected that.

src/main/java/org/dependencytrack/persistence/WorkflowStateQueryManager.java
```diff
@@ -0,0 +1,121 @@
...
+            }
+
+            preparedStatement.execute();
+            rs = preparedStatement.getResultSet();
+            while (rs.next()) {
+                WorkflowState workflowState = new WorkflowState();
+                workflowState.setId(rs.getLong(1));
+                WorkflowState parent = new WorkflowState();
+                parent.setId(rs.getLong(2));
+                workflowState.setParent(parent);
```
nscuro, line 102: Hmmm... I wonder if this can become confusing when `parent` is only ever populated with an ID. Normally `QueryManager`s return persistent objects that would lazy-load such relationships, but that is not the case here. 

It would be good if this was documented in the javadoc. Alternatively, we could use a dedicated projection class that holds a `parentId` instead of a `parent` object.

src/main/java/org/dependencytrack/persistence/WorkflowStateQueryManager.java
```diff
@@ -0,0 +1,201 @@
...
+                    "WHERE \"ID\" IN \n" +
+                    " (WITH RECURSIVE \"CTE_WORKFLOW_STATE\" (\"ID\") AS \n" +
+                    "       (SELECT \"ID\" \n" +
+                    "        FROM public.\"WORKFLOW_STATE\"\n" +
+                    "        WHERE \"PARENT_STEP_ID\" = " +"\'" + parentWorkflowState.getId() +"\'" +"\n" +
+                    "          AND \"TOKEN\" = "  +"\'" + parentWorkflowState.getToken() +"\'" + "\n" +
+                    "        UNION ALL SELECT e.\"ID\" \n" +
+                    "        FROM \"WORKFLOW_STATE\" e\n" +
+                    "        INNER JOIN \"CTE_WORKFLOW_STATE\" o ON o.\"ID\" = e.\"PARENT_STEP_ID\") SELECT \"ID\"\n" +
+                    "     FROM \"CTE_WORKFLOW_STATE\"); ";
```
nscuro, line 189: I gave the prepared statement variant a spin, and the error I am getting is:

```
General error: "java.lang.IndexOutOfBoundsException: Index 2 out of bounds for length 2"
```

I used this code:

```java
final var ps = connection.prepareStatement("""
        UPDATE "WORKFLOW_STATE"
        SET "STATUS" = ?
        WHERE "ID" IN (
            WITH RECURSIVE "CTE_WORKFLOW_STATE" ("ID") AS (
                SELECT "ID"
                    FROM "WORKFLOW_STATE"
                    WHERE "PARENT_STEP_ID" = ? AND "TOKEN" = ?
                UNION ALL
                SELECT "e"."ID"
                    FROM "WORKFLOW_STATE" AS "e"
                    INNER JOIN "CTE_WORKFLOW_STATE" AS "o" ON "o"."ID" = "e"."PARENT_STEP_ID"
            )
            SELECT "ID"
            FROM "CTE_WORKFLOW_STATE"
        )
        """);
ps.setString(1, transientStatus.name());
ps.setLong(2, parentWorkflowState.getId());
ps.setString(3, parentWorkflowState.getToken().toString());

return ps.executeUpdate();
```

Is that the same issue you were seeing?

If so, I think this may be due to [H2's limited support for CTEs](http://www.h2database.com/html/advanced.html#recursive_queries):

> Limitations: Recursive queries need to be of the type UNION ALL, and the recursion needs to be on the second part of the query. No tables or views with the name of the table expression may exist. Different table expression names need to be used when using multiple distinct table expressions within the same transaction and for the same session. All columns of the table expression are of type VARCHAR, and may need to be cast to the required data type. Views with recursive queries are not supported. Subqueries and INSERT INTO ... FROM with recursive queries are not supported. **Parameters are only supported within the last SELECT statement** (a workaround is to use session variables like `@start` within the table expression).

VithikaS: Yes, same issue.

src/main/java/org/dependencytrack/persistence/WorkflowStateQueryManager.java
```diff
@@ -0,0 +1,201 @@
...
+            "   FROM WORKFLOW_STATE e\n" +
+            "   INNER JOIN CTE_WORKFLOW_STATE o ON o.ID = e.PARENT_STEP_ID)\n" +
+            "SELECT ID,\n" +
+            "       PARENT_STEP_ID,\n" +
+            "       STATUS,\n" +
+            "       STEP,\n" +
+            "       TOKEN,\n" +
+            "       STARTED_AT,\n" +
+            "       UPDATED_AT\n" +
+            "FROM CTE_WORKFLOW_STATE";
```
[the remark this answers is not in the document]
VithikaS: yes, much readable now, thanks

nscuro (approved):

PR #243, opened 2023-07-21
Add rest endpoint for workflow status

sahibamittal:
### Description

Add a REST endpoint to fetch status of workflow state upon bom upload.

### Addressed Issue

https://github.com/DependencyTrack/hyades/issues/684

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/resources/v1/WorkflowResource.java
+++ b/src/main/java/org/dependencytrack/resources/v1/WorkflowResource.java
@@ -1,0 +1,54 @@
+package org.dependencytrack.resources.v1;
+
+import alpine.common.logging.Logger;
+import alpine.server.auth.PermissionRequired;
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiParam;
+import io.swagger.annotations.ApiResponse;
+import io.swagger.annotations.ApiResponses;
+import io.swagger.annotations.Authorization;
+import org.dependencytrack.auth.Permissions;
+import org.dependencytrack.model.WorkflowState;
+import org.dependencytrack.persistence.QueryManager;
+import org.dependencytrack.resources.v1.vo.IsTokenBeingProcessedResponse;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.util.List;
+import java.util.UUID;
+
+@Path("/v1/workflow")
+@Api(value = "workflow", authorizations = @Authorization(value = "X-Api-Key"))
+public class WorkflowResource {
+
+    private static final Logger LOGGER = Logger.getLogger(WorkflowResource.class);
+
+    @GET
+    @Path("/token/{uuid}/status")
+    @Produces(MediaType.APPLICATION_JSON)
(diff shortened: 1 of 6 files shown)


src/main/java/org/dependencytrack/resources/v1/WorkflowResource.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @GET
+    @Path("/token/{uuid}/status")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "Retrieves workflow states associated with the token received from bom upload .", response = IsTokenBeingProcessedResponse.class)
+    @ApiResponses(value = {
+            @ApiResponse(code = 401, message = "Unauthorized")
+    })
+    @PermissionRequired(Permissions.Constants.BOM_UPLOAD)
+    public Response isTokenBeingProcessed(
```
nscuro, line 39: Should be renamed to describe more accurately what it's doing.

```suggestion
    public Response getWorkflowStates(
```

src/test/java/org/dependencytrack/resources/v1/WorkflowResourceTest.java
```diff
@@ -0,0 +1,98 @@
...
+        Response response = target(V1_WORKFLOW + "/token/" + uuid + "/status").request()
+                .header(X_API_KEY, apiKey)
+                .get(Response.class);
+
+        assertThat(response.getStatus()).isEqualTo(HttpStatus.SC_OK);
+        JsonArray json = parseJsonArray(response);
+        assertThat(json).isNotNull();
+        assertThat(json.size()).isEqualTo(2);
+        var result = json.getJsonObject(0);
+        assertThat(result.getInt("id")).isEqualTo(1);
```
nscuro, line 69: Database IDs should not be surfaced in the API. We'll need a `@JsonIgnore` annotation on `WorkflowState`'s `id` field.

src/test/java/org/dependencytrack/resources/v1/WorkflowResourceTest.java
```diff
@@ -0,0 +1,98 @@
...
+        assertThat(result.getString("token")).isEqualTo(uuid.toString());
+        assertThat(result.getString("step")).isEqualTo("BOM_CONSUMPTION");
+        assertThat(result.getString("status")).isEqualTo("COMPLETED");
+        result = json.getJsonObject(1);
+        assertThat(result.getInt("id")).isEqualTo(2);
+        assertThat(result.getString("token")).isEqualTo(uuid.toString());
+        assertThat(result.getString("step")).isEqualTo("BOM_PROCESSING");
+        assertThat(result.getString("status")).isEqualTo("PENDING");
```
nscuro, line 77: Just a side note: I think it is better to assert JSON responses using [json-unit](https://github.com/lukas-krecan/JsonUnit#assertj), like so:

https://github.com/DependencyTrack/hyades-apiserver/blob/814cbbbe94bfc1c9aa2d1606400d0abe77e08fff/src/test/java/org/dependencytrack/resources/v1/BomResourceTest.java#L164-L183

That way it is easier to see the actual response format, and you are forced to assert all fields of the response, not only a selected few.

Not a hard requirement, so if you prefer the current way that's fine too.

nscuro (changes requested):
The logic for the existing token status endpoint also needs to be updated:

https://github.com/DependencyTrack/hyades-apiserver/blob/814cbbbe94bfc1c9aa2d1606400d0abe77e08fff/src/main/java/org/dependencytrack/resources/v1/BomResource.java#L342-L360

But IMO it's fine to do that in a separate PR.

nscuro (approved):

PR #247, opened 2023-07-21
Bom consumption and processing workflow

VithikaS:
### Description

This PR is to update workflow states when Bom is uploaded

### Addressed Issue

Addresses issue https://github.com/DependencyTrack/hyades/issues/685

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/WorkflowStateQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/WorkflowStateQueryManager.java
@@ -13,6 +13,7 @@ import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.List;
 import java.util.UUID;
 
@@ -61,7 +62,8 @@ public class WorkflowStateQueryManager extends QueryManager implements IQueryMan
     public static final String UPDATE_WORKFLOW_STATES_QUERY = """
             
             UPDATE "WORKFLOW_STATE"
-            SET "STATUS" = ?
+            SET "STATUS" = ?,
+            "UPDATED_AT" = ?
             WHERE "ID" IN
                 (WITH RECURSIVE "CTE_WORKFLOW_STATE" ("ID") AS
                    (SELECT "ID"
@@ -113,8 +115,8 @@ public class WorkflowStateQueryManager extends QueryManager implements IQueryMan
 
     public WorkflowState getWorkflowStateByTokenAndStep(UUID token, WorkflowStep step) {
         final Query<WorkflowState> query = pm.newQuery(WorkflowState.class, "this.token == :token && this.step == step");
-        query.setParameters(token, step);
-        return query.executeUnique();
+        query.setRange(0, 1);
+        return singleResult(query.execute(token, step));
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/main/java/org/dependencytrack/resources/v1/BomResource.java
```diff
@@ -463,4 +471,25 @@ private File validateAndStoreBom(final byte[] bomBytes, final Project project) t
...
+        processingState.setStep(WorkflowStep.BOM_PROCESSING);
+        processingState.setStatus(WorkflowStatus.PENDING);
+        WorkflowState processingParent = qm.persist(processingState);
+
+        WorkflowState vulnAnalysisState = new WorkflowState();
+        vulnAnalysisState.setParent(processingParent);
+        vulnAnalysisState.setToken(token);
+        vulnAnalysisState.setStep(WorkflowStep.VULN_ANALYSIS);
+        vulnAnalysisState.setStatus(WorkflowStatus.PENDING);
+        qm.persist(vulnAnalysisState);
```
nscuro, line 493: Would be good to have this done in a single DB transaction, so that either all steps are created, or none.

src/main/java/org/dependencytrack/tasks/BomUploadProcessingTask.java
```diff
@@ -158,6 +164,19 @@ public void inform(final Event e) {
...
+        WorkflowState bomConsumptionState = null;
+        List<Component> components = null;
+        List<ServiceComponent> services = null;
+        try (final var qm = new QueryManager()) {
+            WorkflowState consumptionState = qm.getWorkflowStateByTokenAndStep(ctx.uploadToken, WorkflowStep.BOM_CONSUMPTION);
+            if (consumptionState != null) {
+                consumptionState.setStartedAt(Date.from(Instant.now()));
+                bomConsumptionState = qm.persist(consumptionState);
+            } else {
+                LOGGER.warn("Workflow state not found in database so cannot be updated");
```
nscuro, line 176: Including `ctx` in this log would be helpful for correlation.

We should also add a `TODO` here to change this to `error`, and potential throw an exception once the workflow has been migrated completely. Eventually, `BomUploadProcessingTask` should never encounter an event for which no workflow exists.

src/main/java/org/dependencytrack/persistence/WorkflowStateQueryManager.java
```diff
@@ -113,8 +115,8 @@ public List<WorkflowState> getAllWorkflowStatesForAToken(UUID token) {
 
     public WorkflowState getWorkflowStateByTokenAndStep(UUID token, WorkflowStep step) {
         final Query<WorkflowState> query = pm.newQuery(WorkflowState.class, "this.token == :token && this.step == step");
-        query.setParameters(token, step);
-        return query.executeUnique();
+        query.setRange(0, 1);
+        return singleResult(query.execute(token, step));
```
nscuro, line 119: Curious what the reason for this change is. Are we getting more than one result for this query?

nscuro (commented):
Looks good so far and works great! Toggling `BOM_PROCESSING` to `COMPLETED` and some tests are missing, but I understand you already said this was still pending. I'll just flick this PR to draft again.

src/main/java/org/dependencytrack/persistence/WorkflowStateQueryManager.java
VithikaS: There is a constraint on db for uniqueness of token and step but with executeUnique i was getting an exception.

VithikaS: Not required, i have converted it back to executeUnique
There is a uniqueness constraint on db on token and step so we should have only one result

nscuro (approved):

PR #266, opened 2023-08-09
BOM upload processing tweaks

nscuro:
### Description

* Don't log warning when rolling back BOM processing transaction
   * A rollback is never performed automatically by DataNucleus, so this message will always be logged when the transaction failed. We already log the actual cause as error, so this rollback message is just producing noise.
* Avoid redundant copies of uploaded BOM content in-memory
   * The `BomParserFactory#looksLikeCycloneDX` accepts `byte[]`, and looks for certain CycloneDX markers in the provided value. It does so by converting the byte array to `String`, which internally will copy the entire array's content. We already validate the file anyway.

### Addressed Issue

N/A

### Additional Details

N/A

### Checklist


mehab (approved):
lgtm

PR #268, opened 2023-08-11
extend lock only when it is about to expire

VithikaS:
### Description

Lock for extendable tasks is being released as soon as task finishes. This will cause execution on 2 pods if tasks are "too fast".

PR will extend lock only when it is about to expire and it sets the least duration again

### Checklist


sahibamittal (approved):
LGTM

mehab (approved):
lgtm

PR #277, opened 2023-08-18
Fix various bugs in parsing of mirrored vulnerabilities

nscuro:
### Description

* If a vulnerability rating contained only a severity, but no method, score, or vector, the severity was not applied. This is now fixed by applying such severities only when no better option (CVSSv2, CVSSv3, OWASP RR) rating could be applied.
* Vers range parsing failed to correctly parse ranges like `vers:nuget/< 4.3.1`. The splitting logic (https://github.com/DependencyTrack/hyades-apiserver/blob/dd9b8ed2a3753aa7ddb6f6875f56a8cbab4efa51/src/main/java/org/dependencytrack/event/kafka/processor/MirrorVulnerabilityProcessor.java#L179-L183) would break such ranges up into the parts `<` and `4.3.1`, whereas it should've been just `<4.3.1` instead.
* Log messages informing about failures during processing now include the vulnerability ID, to give a little more context.
* Vers ranges are now parsed according to the spec; The implementation for this has been extracted and simplified from https://github.com/nscuro/versatile.
* `MirrorVulnerabilityProcessorTest` now asserts *all* fields of ingested vulnerabilities.

### Addressed Issue

Fixes https://github.com/DependencyTrack/hyades/issues/732

Potentially fixes https://github.com/DependencyTrack/hyades/issues/731 (need to verify)

### Additional Details

This PR relies on the mirror-service changes from https://github.com/DependencyTrack/hyades/pull/756

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/event/kafka/processor/MirrorVulnerabilityProcessor.java
+++ b/src/main/java/org/dependencytrack/event/kafka/processor/MirrorVulnerabilityProcessor.java
@@ -239,19 +239,19 @@ public class MirrorVulnerabilityProcessor implements Processor<String, Bom, Void
                 case GREATER_THAN_OR_EQUAL -> versionStartIncluding = constraint.version();
                 case LESS_THAN_OR_EQUAL -> versionEndIncluding = constraint.version();
                 case LESS_THAN -> versionEndExcluding = constraint.version();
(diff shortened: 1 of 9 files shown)


nscuro:
TODO: Ingest wildcard version ranges

PR #282, opened 2023-08-22
Ensure drainage of internal event system queue, Kafka Streams, and Kafka producer on shutdown

nscuro:
### Description

Previously, upon graceful shutdown, the `ExecutorService`s powering the internal event system were shut down, but there was no waiting for all queued events to be processed. Which meant that in worst case, the API server would accept BOM uploads, queue their processing, and then get shut down before processing could complete.

This is now prevented by first waiting for the event queue to be drained before considering the event system to be terminated. During shutdown, newly dispatched events will not be accepted.

Kafka Streams is being drained and shut down prior to the internal event system. This means that even internal events triggered by KS processing (e.g. policy evaluation, metrics updates) will be processed prior to terminating the application.

The dedicated Kafka producer is shut down after the internal event system, which again means that Kafka messages emitted by internal event processing will be dispatched to Kafka prior to application termination.

Default drain timeouts are sufficiently long to cater to slow processing times, but are configurable if needed.

Classes in `alpine.*` packages are currently shadowed, but those changes will be contributed to Alpine, so this will be temporary.

### Addressed Issue

N/A

### Additional Details

N/A

### Checklist


nscuro:
The Alpine changes were merged to the upstream repository (https://github.com/stevespringett/Alpine/pull/508), but we can't use that yet until a new Alpine version is released.

VithikaS (approved):

PR #291, opened 2023-08-30
Ensure Kafka producer exceptions are always logged

nscuro:
### Description

When we dispatch large quantities of events to Kafka, we do it asynchronously to make use of the producer's batching behavior.

By producing asynchronously, the calling code does not get feedback about whether the production succeeded or not.

The producer's built in retry behavior will kick in nonetheless. For non-retryable errors though, we want to be informed. Per default, the producer does not log such errors, but relies on user-provided `Callback`s to do it instead.

### Addressed Issue

N/A

### Additional Details

N/A

### Checklist


VithikaS (approved):

mehab (approved):
lgtm

PR #311, opened 2023-09-12
Remove `RequirementsVerifier`

nscuro:
### Description

This removes the `RequirementsVerifier`, which prevents startup when the application is presented with fewer resources than the [minimum requirements](https://docs.dependencytrack.org/getting-started/deploy-docker/).

The API server is now able to run on much fewer resources than the vanilla version. Refusing startup when presented with "too few" resources is not necessary anymore.

### Addressed Issue

N/A

### Additional Details

N/A

### Checklist


VithikaS (approved):

VithikaS (commented):
We can remove property from app properties file associated with system requirements check

nscuro:
@VithikaS Ha, good spot! Removed it, and also the `RequirementsException` class that is no longer needed.

In fact, I found that none of the classes in the `exception` package are actually used anywhere. So more code to delete, yay!

VithikaS (approved):

PR #332, opened 2023-09-25
Add license condition test

VithikaS:
### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/policy/cel/CelPolicyEngine.java
+++ b/src/main/java/org/dependencytrack/policy/cel/CelPolicyEngine.java
@@ -15,7 +15,6 @@ import org.apache.commons.collections4.multimap.ArrayListValuedHashMap;
 import org.apache.commons.collections4.multimap.HashSetValuedHashMap;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
-import org.dependencytrack.model.Component;
 import org.dependencytrack.model.Policy;
 import org.dependencytrack.model.PolicyCondition;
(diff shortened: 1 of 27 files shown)


src/main/java/org/dependencytrack/policy/cel/CelPolicyEngine.java
```diff
@@ -157,7 +157,7 @@ public void evaluateProject(final UUID uuid) {
 
             // Preload licenses for the entire project, as chances are high that they will be used by multiple components.
             final Map<Long, org.hyades.proto.policy.v1.License> licenseById;
-            if (requirements.containsKey(TYPE_LICENSE)) {
+            if (requirements.containsKey(TYPE_LICENSE) || requirements.containsKey(TYPE_COMPONENT)) {
```
nscuro, line 160: While adding this check resolves the issue, it will cause licenses to always be loaded, which is not necessarily what we want.

If you set a breakpoint here:

https://github.com/DependencyTrack/hyades-apiserver/blob/777c9f170c849d7071b721ec0258122b668e0708/src/main/java/org/dependencytrack/policy/cel/CelPolicyScriptHost.java#L135-L136

You'll see that we correctly identified that `resolved_license` is being accessed. But because this is an existence check (`has(component.resolved_license)`), the `License` type itself is not present in `requirements`:

image

Instead of checking on the `Component` type, I'd suggest to check for `Component.resolved_license` being accessed:

```suggestion
            if (requirements.containsKey(TYPE_LICENSE) || (requirements.containsKey(TYPE_COMPONENT) && requirements.get(TYPE_COMPONENT).contains("resolved_license"))) {
```

nscuro (approved):

PR #334, opened 2023-09-26
Add initialiser for integrity meta component sync

sahibamittal:
### Description

Create Initialiser which will kick off update of new table with hash information and publishedAt date on application startup.

### Addressed Issue

https://github.com/DependencyTrack/hyades/issues/699

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/event/IntegrityMetaInitializer.java
+++ b/src/main/java/org/dependencytrack/event/IntegrityMetaInitializer.java
@@ -1,0 +1,25 @@
+package org.dependencytrack.event;
+
+import alpine.common.logging.Logger;
+import org.dependencytrack.persistence.QueryManager;
+
+import javax.servlet.ServletContextEvent;
+import javax.servlet.ServletContextListener;
+
+public class IntegrityMetaInitializer implements ServletContextListener {
+
+    private static final Logger LOGGER = Logger.getLogger(IntegrityMetaInitializer.class);
+
+    @Override
+    public void contextInitialized(final ServletContextEvent event) {
+        LOGGER.info("Initializing integrity meta component sync");
+        try (final var qm = new QueryManager()) {
+            // Sync purls from Component only if IntegrityMetaComponent is empty
+            if (qm.getIntegrityMetaComponentCount() == 0) {
+                qm.synchronizeIntegrityMetaComponent();
+            } else {
+                LOGGER.info("Skipping initial integrity meta component synchronizing as data already exists.");
+            }
+        }
+    }
+}
--- a/src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
@@ -785,4 +785,14 @@ final class ComponentQueryManager extends QueryManager implements IQueryManager 
             DbUtil.close(connection);
         }
     }
+
+    /**
+     * Returns the count of records in IntegrityMetaComponent.
+     *
+     * @return the count of records
+     */
+    public int getIntegrityMetaComponentCount() {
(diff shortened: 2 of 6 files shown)


src/main/java/org/dependencytrack/event/IntegrityMetaInitializer.java
```diff
@@ -0,0 +1,89 @@
...
+                query.setParameters(purl);
+                query.setResult("DISTINCT purlCoordinates, internal");
+                ComponentProjection componentProjection = query.executeResultUnique(ComponentProjection.class);
+                kafkaEventDispatcher.dispatchAsync(new ComponentRepositoryMetaAnalysisEvent(componentProjection.purlCoordinates, componentProjection.internal));
+            }
+        }
+    }
+
+    private List<String> fetchNextPurlsPage(PersistenceManager pm, long offset) throws Exception {
+        try (final Query<IntegrityMetaComponent> query = pm.newQuery(IntegrityMetaComponent.class)) {
```
VithikaS, line 80: When we are fetching purls, we should check fetch status and last fetch. We want to avoid picking purls which have  just been sent for analysis

src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
```diff
@@ -785,4 +783,31 @@ public synchronized void synchronizeIntegrityMetaComponent() {
             DbUtil.close(connection);
         }
     }
+
+    /**
+     * Returns the count of records in IntegrityMetaComponent.
+     *
+     * @return the count of records
+     */
+    public int getIntegrityMetaComponentCount() {
```
VithikaS, line 792: Aggregate function is provided by JDO so maybe we can use that here.
Something like this 

    public Long getIntegrityMetaComponentCount() throws Exception {
        try (final Query<IntegrityMetaComponent> query = pm.newQuery(IntegrityMetaComponent.class)) {
            query.setResult("count(this)");
            return query.executeResultUnique(Long.class);
        }
    }

src/main/java/org/dependencytrack/event/IntegrityMetaInitializer.java
```diff
@@ -0,0 +1,89 @@
...
+                final PersistenceManager pm = qm.getPersistenceManager();
+                long offset = 0;
+                long startTime = System.currentTimeMillis();
+                List<String> purls = fetchNextPurlsPage(pm, offset);
+                while (!purls.isEmpty()) {
+                    long cumulativeProcessingTime = System.currentTimeMillis() - startTime;
+                    if(isLockToBeExtended(cumulativeProcessingTime, INTEGRITY_META_INITIALIZER_TASK_LOCK)) {
+                        LockExtender.extendActiveLock(Duration.ofMinutes(5).plus(lockConfiguration.getLockAtLeastFor()), lockConfiguration.getLockAtLeastFor());
+                    }
+                    dispatchPurls(pm, purls);
```
VithikaS, line 57: We should update the last fetch and status of purls dispatched

src/main/java/org/dependencytrack/event/IntegrityMetaInitializer.java
```diff
@@ -0,0 +1,82 @@
...
+    private void process() {
+        LOGGER.info("Initializing integrity meta component sync");
+        try (final var qm = new QueryManager()) {
+            if (qm.getIntegrityMetaComponentCount() == 0) {
+                // Sync purls from Component only if IntegrityMetaComponent is empty
+                qm.synchronizeIntegrityMetaComponent();
+                batchProcessPurls(qm);
+            }
+            // dispatch purls not processed yet
+            batchProcessPurls(qm);
```
VithikaS, line 44: batch processed will be called twice if count is zero

VithikaS (approved):

PR #339, opened 2023-10-02
added integrity analysis event on apiserver

mehab:
### Description
Added integrity analysis event on hyades apiserver side for periodic analysis, manual component creation and bom upload events

### Addressed Issue

Addresses https://github.com/DependencyTrack/hyades/issues/699

### Checklist


src/main/java/org/dependencytrack/resources/v1/ComponentResource.java
```diff
@@ -316,7 +321,15 @@ public Response createComponent(@PathParam("uuid") String uuid, Component jsonCo
             component.setNotes(StringUtils.trimToNull(jsonComponent.getNotes()));
 
             component = qm.createComponent(component, true);
-            kafkaEventDispatcher.dispatchBlocking(new ComponentRepositoryMetaAnalysisEvent(component));
+            ComponentProjection componentProjection =
```
VithikaS, line 324: purl coordinates will not have the qualifier information which will be required for integrity data. We would need the canonicalized form of purl i think.

Comment in Component model:

 `private String purlCoordinates; // Field should contain only type, namespace, name, and version. Everything up to the qualifiers`

src/main/proto/org/hyades/repometaanalysis/v1/repo_meta_analysis.proto
```diff
@@ -11,6 +11,15 @@ option java_package = "org.hyades.proto.repometaanalysis.v1";
 message AnalysisCommand {
   // The component that shall be analyzed.
   Component component = 1;
+  FetchMeta fetch_meta = 2;
+
+}
+
+enum FetchMeta{
+  FETCH_META_UNSPECIFIED = 0;
```
sahibamittal, line 19: Do we need FETCH_META_UNSPECIFIED? as it should be either one or both but never unspecified. If it is unspecified, the event wouldn't be called at all.

mehab: It is part of the proto enum standard. Would not satisfy the enum standard without the unspecified enum field

sahibamittal (approved):
LGTM👍🏼

sahibamittal (approved):

PR #343, opened 2023-10-04
Incoming events with hash information for apiserver

mehab:
added logic for processing incoming hash information as part of repometa analysis

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/event/kafka/processor/RepositoryMetaResultProcessor.java
+++ b/src/main/java/org/dependencytrack/event/kafka/processor/RepositoryMetaResultProcessor.java
@@ -8,6 +8,8 @@ import io.micrometer.core.instrument.Timer;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.kafka.streams.processor.api.Processor;
 import org.apache.kafka.streams.processor.api.Record;
+import org.dependencytrack.model.FetchStatus;
+import org.dependencytrack.model.IntegrityMetaComponent;
 import org.dependencytrack.model.RepositoryMetaComponent;
 import org.dependencytrack.model.RepositoryType;
 import org.dependencytrack.persistence.QueryManager;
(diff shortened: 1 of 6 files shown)


src/main/java/org/dependencytrack/event/kafka/processor/RepositoryMetaResultProcessor.java
```diff
@@ -105,28 +140,48 @@ private void synchronizeRepositoryMetaComponent(final PersistenceManager pm, fin
...
+            LOGGER.warn("""
+                    Received hash information for %s that has already been processed; Discarding
+                    """.formatted(purl));
+            return null;
+        }
+        if (result.hasIntegrityMeta()) {
+            if (result.getIntegrityMeta().hasMd5() || result.getIntegrityMeta().hasSha1() || result.getIntegrityMeta().hasSha256()
+                    || result.getIntegrityMeta().hasSha512() || result.getIntegrityMeta().hasCurrentVersionPublished()) {
+                persistentIntegrityMetaComponent.setMd5(result.getIntegrityMeta().getMd5());
+                persistentIntegrityMetaComponent.setSha256(result.getIntegrityMeta().getSha256());
```
sahibamittal, line 166: Would result.getIntegrityMeta().getSha256() throw exception if integrityMeta objects md5 but doesn't have sha256 since if-condition passed in this case?

src/main/proto/org/hyades/repometaanalysis/v1/repo_meta_analysis.proto
```diff
@@ -25,6 +25,17 @@ message AnalysisResult {
...
+}
+
+message IntegrityMeta {
+  optional string md5 = 1;
+  optional string sha1 = 2;
+  optional string sha256 = 3;
+  optional string sha512 = 4;
+  string integrity_meta_source_url = 5;
+  // When the component current version last modified.
+  optional google.protobuf.Timestamp current_version_published = 6;
```
sahibamittal, line 38: Need to update these to the latest fields
`message IntegrityMeta {
  optional string md5 = 1;
  optional string sha1 = 2;
  optional string sha256 = 3;
  optional string sha512 = 4;
  // When the component current version last modified.
  optional google.protobuf.Timestamp current_version_last_modified = 5;
  // Complete URL to fetch integrity metadata of the component.
  optional string repository_url = 6;
}`

mehab: done :)

src/main/java/org/dependencytrack/event/kafka/processor/RepositoryMetaResultProcessor.java
mehab: done :)

VithikaS (approved):

PR #391, opened 2023-10-24
Fix integrity meta initializer

VithikaS:
### Description

This PR fixes integrity meta initializer
It will select purls from component table and insert into inetgrity meta component table on start up it integrity meta initializer is enabled. It will ignore any conflicts. Check on count has been removed because it is ignoring conflicts and perfroms this activity only on application start up.

Sending events to kafka topic for purls which have not been processed will happen on a scheduled task. This is for existing sboms and components. When sbom is uploaded, it will send events for integrity metadata normally.

### Checklist


mehab (approved):
lgtm

PR #423, opened 2023-11-02
Diff changes made by scanners to existing vulnerabilities

nscuro:
### Description

Instead of only logging *that* they were changed, also log which field was updated, and from which to which value.

Also reduce log level from `WARN` to `INFO`.

This is more or less a good-enough temporary solution, I realize we'll want similar functionality in various places throughout the code base.

image

### Addressed Issue

N/A

### Additional Details

N/A

### Checklist


sahibamittal (approved):
Lgtm

PR #426, opened 2023-11-02
Added id column to sorting

VithikaS:
### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/QueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/QueryManager.java
@@ -213,7 +213,7 @@ public class QueryManager extends AlpineQueryManager {
                 }
             }
             if (found) {
-                query.setOrdering(orderBy + " " + orderDirection.name().toLowerCase() + " " + "id asc");
+                query.setOrdering(orderBy + " " + orderDirection.name().toLowerCase() + ", id asc");
             }
         }
         return query;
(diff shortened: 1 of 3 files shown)


src/main/java/org/dependencytrack/persistence/QueryManager.java
```diff
@@ -189,6 +192,33 @@ public QueryManager(final PersistenceManager pm, final AlpineRequest request) {
...
+            final org.datanucleus.store.query.Query iq = ((JDOQuery) query).getInternalQuery();
+            final String candidateField = orderBy.contains(".") ? orderBy.substring(0, orderBy.indexOf('.')) : orderBy;
+            for (final Field field: iq.getCandidateClass().getDeclaredFields()) {
+                if (candidateField.equals(field.getName())) {
+                    found = true;
+                    break;
+                }
+            }
+            if (found) {
+                query.setOrdering(orderBy + " " + orderDirection.name().toLowerCase() + " " + "id asc");
```
nscuro, line 216: I think there is a comma missing here:

```suggestion
                query.setOrdering(orderBy + " " + orderDirection.name().toLowerCase() + ", id asc");
```

VithikaS: It was! Interestingly, there was no exception, just ordering part was truncated when it was added without comma

PR #428, opened 2023-11-03
Fix component published date not being returned when no integrity analysis exists for it

VithikaS:
### Description

Query to fetch integrity metadata should perform a left join on integrity analysis table. We could be having "published_at" data but not integrity analysis data

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/QueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/QueryManager.java
@@ -1894,15 +1894,17 @@ public class QueryManager extends AlpineQueryManager {
             if (resultSet.next()) {
                 Date publishedDate = null;
                 Date lastFetch = null;
+                IntegrityMatchStatus integrityMatchStatus = null;
                 if(resultSet.getTimestamp("PUBLISHED_AT") != null) {
                     publishedDate = Date.from(resultSet.getTimestamp("PUBLISHED_AT").toInstant());
                 }
                 if(resultSet.getTimestamp("LAST_FETCH") != null) {
                     lastFetch = Date.from(resultSet.getTimestamp("LAST_FETCH").toInstant());
                 }
-                return new ComponentMetaInformation(publishedDate,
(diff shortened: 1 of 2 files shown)


src/main/java/org/dependencytrack/persistence/QueryManager.java
```diff
@@ -1883,7 +1883,7 @@ public ComponentMetaInformation getMetaInformation(UUID uuid) {
         PreparedStatement preparedStatement = null;
         String queryString = """
                 SELECT "C"."ID", "C"."PURL", "IMC"."LAST_FETCH",  "IMC"."PUBLISHED_AT", "IA"."INTEGRITY_CHECK_STATUS" FROM "COMPONENT" "C"
-                JOIN "INTEGRITY_META_COMPONENT" "IMC" ON "C"."PURL" ="IMC"."PURL" JOIN "INTEGRITY_ANALYSIS" "IA" ON "IA"."COMPONENT_ID" ="C"."ID"  WHERE "C"."UUID" = ?
+                JOIN "INTEGRITY_META_COMPONENT" "IMC" ON "C"."PURL" ="IMC"."PURL" LEFT JOIN "INTEGRITY_ANALYSIS" "IA" ON "IA"."COMPONENT_ID" ="C"."ID"  WHERE "C"."UUID" = ?
```
nscuro, line 1886: We now also need a `null` check here:

https://github.com/DependencyTrack/hyades-apiserver/blob/8bb3729b90da11abf9f2c4eaec2005f15409276d/src/main/java/org/dependencytrack/persistence/QueryManager.java#L1903-L1905

`IntegrityMatchStatus.valueOf(resultSet.getString("INTEGRITY_CHECK_STATUS"))` will throw when no `INTEGRITY_CHECK_STATUS` is returned.

VithikaS: yes, added.

nscuro (approved):

PR #441, opened 2023-11-16
Convert Jdo to Sql query to fix sorting on components fields

sahibamittal:
### Description

Convert Jdo to Sql query to fix sorting on components fields 'Published_At' and integrity status.

### Checklist


src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
```diff
@@ -156,56 +159,177 @@ public PaginatedResult getComponents(final Project project, final boolean includ
...
+            }
+            if (orderBy.equalsIgnoreCase("componentMetaInformation.integrityMatchStatus")) {
+                queryString +=
+                        """
+                            ORDER BY "INTEGRITY_CHECK_STATUS"
+                        """;
+            }
+            if (queryString.contains("ORDER BY")) {
+                if (orderDirection == OrderDirection.ASCENDING) {
+                    queryString += " ASC ";
```
VithikaS, line 291: I think we should add id field in the ordering otherwise it might re-introduce the bug of duplicates on a page when we are sorting on nullable fields

src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
```diff
@@ -156,56 +159,177 @@ public PaginatedResult getComponents(final Project project, final boolean includ
...
     public PaginatedResult getComponents(final Project project, final boolean includeMetrics, final boolean onlyOutdated, final boolean onlyDirect) {
-        final PaginatedResult result;
-        String querySring = "SELECT FROM org.dependencytrack.model.Component WHERE project == :project ";
-        if (filter != null) {
-            querySring += " && (project == :project) && name.toLowerCase().matches(:name)";
-        }
+        List<Component> componentsResult = new ArrayList<>();
+        final var params = new HashMap<String, Object>();
+        String queryString = """
+                        SELECT 'org.dependencytrack.model.Component' AS "DN_TYPE",
```
VithikaS, line 165: If we are using native query, selecting as DN_TYPE would not be required

src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
```diff
@@ -156,56 +159,178 @@ public PaginatedResult getComponents(final Project project, final boolean includ
...
                     final RepositoryType type = RepositoryType.resolve(purl);
                     if (RepositoryType.UNSUPPORTED != type) {
                         final RepositoryMetaComponent repoMetaComponent = getRepositoryMetaComponent(type, purl.getNamespace(), purl.getName());
                         component.setRepositoryMeta(repoMetaComponent);
-                        component.setComponentMetaInformation(getMetaInformation(component.getUuid()));
                     }
                 }
             }
+            componentsResult.add(component);
         }
```
nscuro, line 332: Can't we populate this from the SQL query's result as well?

src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
```diff
@@ -156,56 +159,178 @@ public PaginatedResult getComponents(final Project project, final boolean includ
...
+                        """
+                            ORDER BY "NAME"
+                        """;
+            }
+            if (orderBy.equalsIgnoreCase("componentMetaInformation.integrityMatchStatus")) {
+                queryString +=
+                        """
+                            ORDER BY "INTEGRITY_CHECK_STATUS"
+                        """;
+            }
```
nscuro, line 286: In the UI it is also (currently) possible to sort by:

* Group
* Risk score

So at least those need to be added here as well.

src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
```diff
@@ -156,56 +159,178 @@ public PaginatedResult getComponents(final Project project, final boolean includ
...
-                            " SELECT FROM org.dependencytrack.model.RepositoryMetaComponent m " +
-                            " WHERE m.name == this.name " +
-                            " && m.namespace == this.group " +
-                            " && m.latestVersion != this.version " +
-                            " && this.purl.matches('pkg:' + m.repositoryType.toString().toLowerCase() + '/%') " +
-                            " ).isEmpty()";
+            queryString +=
+                    """
+                        AND NOT (NOT EXISTS (
+                        SELECT "M"."ID" AS "DN_APPID"
```
nscuro, line 243: `"M"."ID" AS "DN_APPID"` appears to be from the DataNucleus-generated query still, should be safe to remove (or rename, `DN_APPID` doesn't have any meaning).

src/main/java/org/dependencytrack/util/ComponentUtil.java
```diff
@@ -0,0 +1,130 @@
...
+import java.io.ObjectInputStream;
+import java.util.Collections;
+import java.util.List;
+import java.util.UUID;
+
+public class ComponentUtil {
+
+    private static final Logger LOGGER = Logger.getLogger(ComponentUtil.class);
+
+    public static final Component mapToComponent(ComponentProjection result) {
```
nscuro, line 24: This method could be moved to the `ComponentProjection` class.

src/main/java/org/dependencytrack/model/sqlMapping/ComponentProjection.java
```diff
@@ -0,0 +1,199 @@
+package org.dependencytrack.model.sqlMapping;
```
nscuro, line 1: Please keep package names in lower case.

src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
sahibamittal: There are two separate tables 'DependencyMetrics' and 'Repository_meta_component' being queried in this. In order to populate these objects from same SQL query we can add more joins on these 2 tables.

sahibamittal: Not including DependencyMetrics query join, it'll complicate the query and make it less readable/maintainable.
Querying repo_meta_component is problematic as well because it'll need 3 fields (from parsing a purl) -> name, namespace and repository_type which is an enum. We wont have this enum value while building the query, it is calculated separately depending on component purl.

src/main/java/org/dependencytrack/model/sqlmapping/ComponentProjection.java
```diff
@@ -0,0 +1,245 @@
...
+
+    public Boolean isOsiApproved;
+
+    public Boolean isFsfLibre;
+
+    public Boolean isCustomLicense;
+
+    public Long totalCount;
+
+    public static final Component mapToComponent(ComponentProjection result) {
```
mehab, line 145: this method can be non final

src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
```diff
@@ -156,56 +160,191 @@ public PaginatedResult getComponents(final Project project, final boolean includ
...
-            final String filterString = ".*" + filter.toLowerCase() + ".*";
-            result = execute(query, project, filterString);
+            queryString +=
+                    """
+                        ORDER BY "name",
+                        "version" DESC
+                    """;
         } else {
-            result = execute(query, project);
+            if (orderBy.equalsIgnoreCase("componentMetaInformation.publishedDate")) {
```
mehab, line 265: would it be more readable to use switch expression on the orderBy value and then have the ORDER BY part appended to the query with the final result?

mehab (approved):
lgtm

PR #455, opened 2023-11-23
Add `vers` range validation before compilation

sahibamittal:
### Description

It would be good if we could perform the validation during the CEL script compilation, in order to give users feedback synchronously.

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/873

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/policy/cel/CelPolicyScriptHost.java
+++ b/src/main/java/org/dependencytrack/policy/cel/CelPolicyScriptHost.java
@@ -118,7 +118,14 @@ public class CelPolicyScriptHost {
 
             final Ast ast = astIssuesTuple.getAst();
             final Program program = environment.program(ast);
-            final MultiValuedMap<Type, String> requirements = analyzeRequirements(CEL.astToCheckedExpr(ast));
+            final var expr = CEL.astToCheckedExpr(ast);
+            final MultiValuedMap<Type, String> requirements = analyzeRequirements(expr);
+
+            // perform vers range validity
+            if (requirements.get(TYPE_PROJECT).contains("version") || requirements.get(TYPE_COMPONENT).contains("version")) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/dependencytrack/policy/cel/CelPolicyScriptHost.java
```diff
@@ -118,7 +118,14 @@ public CelPolicyScript compile(final String scriptSrc, final CacheMode cacheMode
...
             final Ast ast = astIssuesTuple.getAst();
             final Program program = environment.program(ast);
-            final MultiValuedMap<Type, String> requirements = analyzeRequirements(CEL.astToCheckedExpr(ast));
+            final var expr = CEL.astToCheckedExpr(ast);
+            final MultiValuedMap<Type, String> requirements = analyzeRequirements(expr);
+
+            // perform vers range validity
+            if (requirements.get(TYPE_PROJECT).contains("version") || requirements.get(TYPE_COMPONENT).contains("version")) {
+                final var visitor = new CelPolicyScriptVisitor(expr.getTypeMapMap());
+                visitor.visitVersRangeCheck(expr.getExpr());
```
nscuro, line 127: Passing the entire original `Expr` here will only work when it only contains `(component|project).matches_range("...")` . But keep in mind there can be multiple, e.g. 

```py
component.name == "foo" && component.matches_range("...")
```

or even more deeply-nested ones like:

```py
component.name == "foo" 
  || vulns.exists(vuln, vuln.id == "foo" && component.matches_range("..."))
```

Also there can be multiple calls of the function:

```py
project.matches_range("foo")
  && component.matches_range("bar")
```

So before performing this check, you need to find the `CallExpr` for the `matches_vers` function call(s) *within* the overall expression. That is what the visitor is for: You traverse the abstract syntax tree until you find the expression you're looking for.

nscuro (approved):

PR #462, opened 2023-12-04
Create table for vulnerability policy

VithikaS:
### Description

This PR creates table for Vulnerability Policy and provides persistence for vulnerability policy object

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/942

### Checklist

- [x] I have read and understand the [contributing guidelines](../CONTRIBUTING.md#pull-requests)
- [ ] This PR fixes a defect, and I have provided tests to verify that the fix is effective
- [x] This PR implements an enhancement, and I have provided tests to verify that it works as intended
- [ ] This PR introduces changes to the database model, and I have added corresponding [update logic](https://github.com/DependencyTrack/dependency-track/tree/master/src/main/java/org/dependencytrack/upgrade)
- [ ] This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/dependency-track/tree/master/docs/_docs) accordingly

PR #464, opened 2023-12-04
Implement `VulnerabilityPolicyEvaluator` (without dynamic loading of required fields)

nscuro:
### Description

Adds definitions for vulnerability policies, and implements an evaluator powered by CEL.

The evaluator currently does not dynamically load required fields from the database. Because that includes a lot more code, I decided to decouple that from this initial PR to make reviews easier.

There are now two policy types, each with their own context (only difference currently are the available variables).

### Addressed Issue

Partly addresses https://github.com/DependencyTrack/hyades/issues/937

### Additional Details

N/A

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/policy/vulnerability/VulnerabilityPolicy.java
+++ b/src/main/java/org/dependencytrack/policy/vulnerability/VulnerabilityPolicy.java
@@ -1,0 +1,108 @@
+package org.dependencytrack.policy.vulnerability;
+
+import java.time.ZonedDateTime;
+import java.util.List;
+
+public class VulnerabilityPolicy {
+
+    private String name;
+
+    private String description;
+
+    private String author;
+
+    private ZonedDateTime created;
+
+    private ZonedDateTime updated;
+
+    private ZonedDateTime validFrom;
+
+    private ZonedDateTime validUntil;
+
+    private List<String> conditions;
+
+    private VulnerabilityPolicyAnalysis analysis;
+
+    private List<VulnerabilityPolicyRating> ratings;
+
+    public String name() {
+        return name;
+    }
+
+    public void setName(final String name) {
+        this.name = name;
+    }
+
+    public String description() {
+        return description;
+    }
+
+    public void setDescription(final String description) {
+        this.description = description;
+    }
+
+    public String author() {
+        return author;
+    }
+
+    public void setAuthor(final String author) {
+        this.author = author;
+    }
+
+    public ZonedDateTime created() {
+        return created;
+    }
+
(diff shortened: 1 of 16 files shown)


src/main/java/org/dependencytrack/policy/vulnerability/VulnerabilityPolicy.java
```diff
@@ -0,0 +1,108 @@
+package org.dependencytrack.policy.vulnerability;
+
+import java.time.ZonedDateTime;
+import java.util.List;
+
+public class VulnerabilityPolicy {
```
nscuro, line 6: I intentionally duplicated this class from the persistence model introduced in #462. The intention is that the policy evaluator should not have to care where the policies are coming from. In case we want this to be an SPI in the future, it should also not depend on any internal model definitions.

Curious about opinions on these two approaches:

* Implementations of `VulnerabilityPolicyProvider` map their internal model to this `VulnerabilityPolicy` class
* We make `VulnerabilityPolicy` an interface instead, and the internal model of the `VulnerabilityPolicyProvider` simply implements that

The former makes it easier to deal with breaking changes, the latter means that providers do not need to do any mapping. Thoughts?

VithikaS: If I was a provider, I usually get caught up in mapping because of doubts on what certain fields mean. Interface appears cleaner approach. But when we are provider also, first approach could be quickest to implement.

nscuro: I was just about to change it to an interface, but then noticed that by doing it this way, we also have a hard coupling of data types. Say for `validFrom`, the database model uses `java.util.Date`:

https://github.com/DependencyTrack/hyades-apiserver/blob/fd958b2967f0fe1f0b8ae6f5a22da3ff19b8b393/src/main/java/org/dependencytrack/model/VulnerabilityPolicy.java#L64-L66

but the `VulnerabilityPolicy` interface may use the newer `java.time.Instant` instead.

If the interface defines a method `getValidFrom`, then our internal data model would have to implement this method, using the data type dictated by the interface. The ORM and JSON mappers use getters and setters to access fields, so a constellation like:

```java
private Date validFrom;

@Override // Implements VulnerabilityAnalysis interface
public Instant getValidFrom() {}

public void setValidFrom(Date validFrom) {}
```

could be problematic. Just using `Date` and `Instant` as examples, might just as well be other data types. Limitations of the ORM should not dictate what data types we can use in a (potentially) public SPI.

VithikaS (approved):

PR #487, opened 2023-12-13
Add s3 bundle fetch policy support

mehab:
### Description

Added support for s3 policy bundle fetch. Clean up and refactoring of existing code.
Addresses https://github.com/DependencyTrack/hyades/issues/938 as an enhancement

### Checklist


src/main/java/org/dependencytrack/tasks/vulnerabilitypolicy/S3Client.java
```diff
@@ -0,0 +1,57 @@
...
+
+    private static final Logger LOGGER = Logger.getLogger(S3Client.class);
+
+    public S3Client() {
+        this.bundleFetchEndpoint = Config.getInstance().getProperty(ConfigKey.VULNERABILITY_POLICY_BUNDLE_URL);
+        this.bundleToFetch = Config.getInstance().getProperty(ConfigKey.VULNERABILITY_POLICY_S3_BUNDLE_NAME);
+        this.s3AccessKey = Config.getInstance().getProperty(ConfigKey.VULNERABILITY_POLICY_S3_ACCESS_KEY);
+        this.s3SecretKey = Config.getInstance().getProperty(ConfigKey.VULNERABILITY_POLICY_S3_SECRET_KEY);
+        this.s3BucketName = Config.getInstance().getProperty(ConfigKey.VULNERABILITY_POLICY_S3_BUCKET_NAME);
+        this.region = Config.getInstance().getProperty(ConfigKey.VULNERABILITY_POLICY_S3_REGION);
```
nscuro, line 35: I think this class would be easier to test if we'd use a package-private constructor instead, which accepts all of these values as arguments. In the test, we could then simply construct instances with:

```java
new S3Client(myBundleFetchEndpoint, myBundleToFetch, myAccessKey, mySecretKey, myBucketName, myRegion);
```

For the "production" side of things, where the default `public` constructor is used, you can simply do:

```java
public S3Client() {
    this(Config.getInstance().getProperty(ConfigKey.VULNERABILITY_POLICY_BUNDLE_URL),
        Config.getInstance().getProperty(ConfigKey.VULNERABILITY_POLICY_S3_BUNDLE_NAME),
        Config.getInstance().getProperty(ConfigKey.VULNERABILITY_POLICY_S3_ACCESS_KEY),
        Config.getInstance().getProperty(ConfigKey.VULNERABILITY_POLICY_S3_SECRET_KEY),
        Config.getInstance().getProperty(ConfigKey.VULNERABILITY_POLICY_S3_BUCKET_NAME),
        Config.getInstance().getProperty(ConfigKey.VULNERABILITY_POLICY_S3_REGION));
}
```

Fiddling with environment variables in a unit test should be a workaround, not the norm.

src/main/java/org/dependencytrack/tasks/vulnerabilitypolicy/S3Client.java
```diff
@@ -0,0 +1,57 @@
...
+        if (bundleToFetch == null || s3BucketName == null || s3SecretKey == null || s3AccessKey == null) {
+            LOGGER.error("The required parameters for S3 file access have not been provided. Please check");
+        }
```
nscuro, line 38: Why not throw an `IllegalArgumentException` here? The caller would otherwise never know that this client has no way of ever working correctly.

src/main/java/org/dependencytrack/tasks/vulnerabilitypolicy/S3Client.java
```diff
@@ -0,0 +1,57 @@
...
+        this.minioClient = builder.build();
+    }
+
+    public MinioClient getMinioClient() {
+        if (this.minioClient == null) {
+            createClient();
+        }
+        return minioClient;
+    }
```
nscuro, line 55: This is a bit of a leaky abstraction here. Users of the `S3Client` class should ideally never know that it uses MinIO behind the scenes. 

Would it be possible to implement the high-level functionality directly in `S3Client`, instead of handing out instances of `MinioClient`? For example, in `S3StorageHandler`, you do three things:

1. Check if a bucket exists
2. Check if a given object exists within that bucket, and return its ETag if it does
3. Download an object, and return it as byte array

So, theoretically, in `S3Client` you could expose these methods:

```java
public boolean doesBucketExist(String bucket) throws IOException {
    // TODO
}

public String getObjectEtag(String bucket, String objectName) throws IOException {
    // TODO
}

public byte[] getObject(String bucket, String objectName) throws IOException {
    // TODO
}
```

src/main/java/org/dependencytrack/tasks/vulnerabilitypolicy/blobstorage/S3StorageHandler.java
```diff
@@ -0,0 +1,65 @@
...
+import org.dependencytrack.util.VulnerabilityPolicyUtil;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+import java.util.zip.ZipInputStream;
+
+public class S3StorageHandler extends AbstractVulnerabilityPolicyFetcher {
+    public static S3Client S3CLIENT = new S3Client();
```
nscuro, line 27: What's the reasoning behind making this `public static` instead of `private final` (and not `static`)? AFAICT this variable is exclusively used internally in this class.

src/main/java/org/dependencytrack/util/VulnerabilityPolicyUtil.java
```diff
@@ -0,0 +1,169 @@
...
+            }
+
+        });
+    }
+
+    private static JsonSchema enrichYamlOjMapperAndGetSchema() throws IOException {
+        final String jsonSchemaContent = resourceToString("/schema/vulnerability-policy-v1.schema.json", StandardCharsets.UTF_8);
+        JsonSchemaFactory factory = JsonSchemaFactory.builder(JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V202012)).objectMapper(MAPPER).build();
+        return factory.getSchema(jsonSchemaContent);
+    }
```
nscuro, line 165: This is a good example of something that could be made a `private static final` field instead. Construction is kind of expensive (as it involves I/O and file parsing of the schema), and it's used multiple times (once per policy file).

Re-creating this for each policy file is unnecessary overhead and is best avoided.

pom.xml
```diff
@@ -453,6 +455,11 @@
             <artifactId>jackson-dataformat-yaml</artifactId>
             <version>${lib.jackson.version}</version>
         </dependency>
+        <dependency>
+            <groupId>io.minio</groupId>
+            <artifactId>minio</artifactId>
+            <version>8.5.7</version>
```
nscuro, line 461: Nit: Please use a `lib.minio.version` property here to be consistent with how it's done for other dependencies.

src/main/java/org/dependencytrack/tasks/vulnerabilitypolicy/blobstorage/BlobStorageAccessHandler.java
```diff
@@ -0,0 +1,26 @@
...
+
+public interface BlobStorageAccessHandler {
+
+    default boolean verifyDownloadNeeded() throws ServerException, InsufficientDataException, ErrorResponseException, IOException, NoSuchAlgorithmException, InvalidKeyException, InvalidResponseException, XmlParserException, InternalException {
+        return true;
+    }
+
+    default ZipInputStream downloadZippedContent() throws IOException, ServerException, InsufficientDataException, ErrorResponseException, NoSuchAlgorithmException, InvalidKeyException, InvalidResponseException, XmlParserException, InternalException {
+        return null;
+    }
```
nscuro, line 24: These interface methods should not throw MinIO exceptions. Try to reduce it to 1-2 generic exceptions (or create a custom exception class), for example `IOException`. Implementations will then need to re-throw their internal exceptions accordingly.

src/main/java/org/dependencytrack/tasks/vulnerabilitypolicy/AbstractVulnerabilityPolicyFetcher.java
```diff
@@ -1,174 +1,36 @@
...
-import java.sql.Connection;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Set;
-import java.util.zip.ZipEntry;
-import java.util.zip.ZipInputStream;
-
-import static org.apache.commons.io.IOUtils.resourceToString;
 
 public abstract class AbstractVulnerabilityPolicyFetcher implements BlobStorageAccessHandler {
```
nscuro, line 16: This class appears to not fulfill any use anymore. The functionality it has seems to be exclusive to the `NginxStorageHandler`, as `S3StorageHandler` doesn't use / need any of it. Perhaps the remaining code in here can be moved to `NginxStorageHandler`, and this class can be removed?

src/main/java/org/dependencytrack/tasks/vulnerabilitypolicy/blobstorage/NginxStorageHandler.java
```diff
@@ -0,0 +1,51 @@
...
+                Header[] httpHeaders = response.getAllHeaders();
+                Optional<Header> headerOptional = Arrays.stream(httpHeaders).filter(header -> header.getName().equalsIgnoreCase("ETag")).findFirst();
+                if (headerOptional.isPresent()) {
+                    return VulnerabilityPolicyUtil.matchWithHashConfigProperty(headerOptional.get().getValue());
+                } else {
+                    LOGGER.warn("Was not able to find Etag header in the nginx request. Will proceed assuming that file needs to be downloaded");
+                    return true;
+                }
+            } else {
+                throw new IOException("Unable to get response from resource bundle endpoint for policy");
```
nscuro, line 38: Can we provide more information here? The exception is thrown when the response status is not `200`, so what was it *actually*?

src/main/java/org/dependencytrack/tasks/vulnerabilitypolicy/blobstorage/S3StorageHandler.java
```diff
@@ -0,0 +1,65 @@
...
+        MinioClient minioClient = S3CLIENT.getMinioClient();
+
+        if (minioClient.bucketExists(BucketExistsArgs.builder().bucket(S3CLIENT.getS3BucketName()).build())) {
+            //perform head request check on the object of interest
+            StatObjectResponse response = minioClient.statObject(StatObjectArgs.builder().
+                    bucket(S3CLIENT.getS3BucketName()).object(S3CLIENT.getBundleToFetch()).build());
+            return VulnerabilityPolicyUtil.matchWithHashConfigProperty(response.etag());
+        } else {
+            LOGGER.error("Provided bucket %s not found in the s3 bucket".formatted(S3CLIENT.getS3BucketName()));
+            return false;
```
nscuro, line 51: Should this throw an exception (e.g. `NoSuchElementException`) instead?

src/main/java/org/dependencytrack/util/VulnerabilityPolicyUtil.java
```diff
@@ -0,0 +1,169 @@
...
+                }
+            } else {
+                logger.warn("Was not able to create vulnerability policy object successfully");
+            }
+        } else {
+            StringBuilder errorMessage = new StringBuilder();
+            for (var message : validateMsg) {
+                errorMessage.append(message.getMessage()).append("\n");
+            }
+            logger.error("unable to compile cel policy correctly. Errors received:%s".formatted(errorMessage.toString()));
```
nscuro, line 64: Should this throw an exception instead? It should be up to the caller of `parseVulnerabilityPolicy` to decide how to best deal with errors, and this method already has `throws ScriptCreateException, IOException` in its signature.

src/main/java/org/dependencytrack/util/VulnerabilityPolicyUtil.java
```diff
@@ -0,0 +1,169 @@
...
+            validateVulnerabilityCelPolicy(MAPPER.convertValue(jsonNode.get("conditions"), ArrayList.class));
+            VulnerabilityPolicy vulnerabilityPolicy = MAPPER.convertValue(jsonNode, VulnerabilityPolicy.class);
+            if (vulnerabilityPolicy != null) {
+                if (policyNames.contains(vulnerabilityPolicy.getName())) {
+                    updateVulnerabilityPolicyList.add(vulnerabilityPolicy);
+                } else {
+                    createVulnerabilityPolicyList.add(vulnerabilityPolicy);
+                }
+            } else {
+                logger.warn("Was not able to create vulnerability policy object successfully");
```
nscuro, line 57: Same as below, I think this should throw an exception instead and have callers of this method deal with it instead.

src/main/java/org/dependencytrack/tasks/vulnerabilitypolicy/blobstorage/GenericStorageHandler.java
```diff
@@ -0,0 +1,6 @@
+package org.dependencytrack.tasks.vulnerabilitypolicy.blobstorage;
+
+import org.dependencytrack.tasks.vulnerabilitypolicy.AbstractVulnerabilityPolicyFetcher;
+
+public class GenericStorageHandler extends AbstractVulnerabilityPolicyFetcher {
```
nscuro, line 5: Do we need this class? It looks like `downloadZippedContent` always returns `null`, but `verifyDownloadNeeded` always returns `true`. Code using this handler will most definitely break.

nscuro (changes requested):

nscuro (approved):

PR #490, opened 2023-12-15
Add REST endpoint for vulnerability policies

VithikaS:
### Description

This PR is to create an endpoint to get all vulnerability policies for displaying it on UI

Closes https://github.com/DependencyTrack/hyades/issues/961

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/VulnerabilityPolicyQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/VulnerabilityPolicyQueryManager.java
@@ -1,11 +1,14 @@
 package org.dependencytrack.persistence;
 
 import alpine.common.logging.Logger;
+import alpine.persistence.OrderDirection;
+import alpine.persistence.PaginatedResult;
 import alpine.resources.AlpineRequest;
 import alpine.server.util.DbUtil;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.datatype.jsonorg.JsonOrgModule;
+import org.apache.commons.lang3.tuple.Pair;
 import org.dependencytrack.model.VulnerabilityPolicy;
 import org.dependencytrack.policy.vulnerability.VulnerabilityPolicyAnalysis;
 import org.json.JSONArray;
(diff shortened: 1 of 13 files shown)


src/main/java/org/dependencytrack/persistence/VulnerabilityPolicyQueryManager.java
```diff
@@ -229,4 +257,97 @@ public int updateVulnerabilityPolicyByName(VulnerabilityPolicy vulnerabilityPoli
...
+        vulnerabilityPolicy.setAnalysis(MAPPER.convertValue(analysisJson, VulnerabilityPolicyAnalysis.class));
+        vulnerabilityPolicy.setAuthor(rs.getString("AUTHOR"));
+        Array conditions = rs.getArray("CONDITIONS");
+        vulnerabilityPolicy.setConditions((String[]) conditions.getArray());
+        vulnerabilityPolicy.setCreated(rs.getTimestamp("CREATED"));
+        vulnerabilityPolicy.setDescription(rs.getString("DESCRIPTION"));
+        vulnerabilityPolicy.setName(rs.getString("NAME"));
+        JSONArray ratingsJson = new JSONArray(rs.getString("RATINGS"));
+        vulnerabilityPolicy.setRatings(MAPPER.convertValue(ratingsJson, new TypeReference<>() {
+        }));
```
nscuro, line 317: I just stumbled over this in testing: There needs to be a `null` check here as `ratings` are optional.

Also I think you can pass the result of `rs.getString("RATINGS")` directly to `MAPPER.readValue`. No need to convert to `JSONArray` first.

src/test/java/org/dependencytrack/resources/v1/VulnerabilityPolicyResourceTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Override
+    protected DeploymentContext configureDeployment() {
+        return ServletDeploymentContext.forServlet(new ServletContainer(
+                new ResourceConfig(VulnerabilityPolicyResource.class)
+                        .register(ApiFilter.class)
+                        .register(AuthenticationFilter.class)))
+                .build();
+    }
+
```
nscuro, line 51: Can we have another test here that checks whether all policy fields are returned as expected? i.e. using `assertThatJson` like so: https://github.com/DependencyTrack/hyades-apiserver/blob/850a98862dade23fb10fc6db6657f520b5019a48/src/test/java/org/dependencytrack/resources/v1/VexResourceTest.java#L127-L204

VithikaS: Added a test

nscuro (approved):

PR #499, opened 2023-12-19
API endpoint to update vulnerability policy

sahibamittal:
### Description

Adding new API endpoint to enable editing of existing vulnerability policies.

### Addressed Issue

https://github.com/DependencyTrack/hyades/issues/965

### Checklist


src/main/java/org/dependencytrack/policy/vulnerability/VulnerabilityPolicyProvider.java
```diff
@@ -21,4 +21,5 @@ public interface VulnerabilityPolicyProvider {
 
     PaginatedResult getAllVulnerabilityPolicies(AlpineRequest request);
 
+    VulnerabilityPolicy updateVulnerabilityPolicy(final org.dependencytrack.model.VulnerabilityPolicy updatedPolicy);
```
VithikaS, line 24: We should not pass the database object here because any implementation of PolicyProvider should work

PR #516, opened 2024-01-05
Support bearer auth for policy bundle server

mehab:
### Description

Added support for bearer token based authentication for bundle server. 
If etag header is not provided by bundle server in head request then the policy bundle will always be downloaded and the lastModifiedHash config property would be set to null. 
If etag is returned from head request, the system would receive benefit of avoiding redundant downloads when bundle is same

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/common/ConfigKey.java
+++ b/src/main/java/org/dependencytrack/common/ConfigKey.java
@@ -83,7 +83,7 @@ public enum ConfigKey implements Config.Key {
     VULNERABILITY_POLICY_BUNDLE_URL("vulnerability.policy.bundle.url", null),
     VULNERABILITY_POLICY_BUNDLE_SOURCE_TYPE("vulnerability.policy.bundle.source.type", "NGINX"),
     VULNERABILITY_POLICY_BUNDLE_AUTH_USERNAME( "vulnerability.policy.bundle.auth.username", null),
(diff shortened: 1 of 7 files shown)


src/main/java/org/dependencytrack/common/ConfigKey.java
```diff
@@ -83,6 +83,7 @@ public enum ConfigKey implements Config.Key {
     VULNERABILITY_POLICY_BUNDLE_URL("vulnerability.policy.bundle.url", null),
     VULNERABILITY_POLICY_BUNDLE_SOURCE_TYPE("vulnerability.policy.bundle.source.type", "NGINX"),
     VULNERABILITY_POLICY_BUNDLE_AUTH_USERNAME( "vulnerability.policy.bundle.auth.username", null),
+    VULNERABILITY_POLICY_BUNDLE_BEARER_TOKEN("vulnerability.policy.bundle.bearer.token", null),
```
nscuro, line 86: For consistency:

```suggestion
    VULNERABILITY_POLICY_BUNDLE_AUTH_BEARER_TOKEN("vulnerability.policy.bundle.auth.bearer.token", null),
```

nscuro (commented):
Just one minor suggestion regarding the new property name, otherwise LGTM.

nscuro (approved):

PR #540, opened 2024-01-25
Remove legacy policy engine

mehab:
### Description
Removed legacy policy engine. Removed the cel policy engine enabled flag so only the cel policy engine would always be used

Closes https://github.com/DependencyTrack/hyades/issues/1031

### Checklist


nscuro (approved):
🔥

PR #553, opened 2024-02-05
Migrate `MirrorVulnerabilityProcessor` from Kafka Streams to Parallel Consumer

nscuro:
### Description

Migrates `MirrorVulnerabilityProcessor` from Kafka Streams to Parallel Consumer.

> [!WARNING]
> ~Depends on https://github.com/DependencyTrack/hyades-apiserver/pull/552, please review and merge that first!~

### Addressed Issue

Relates to https://github.com/DependencyTrack/hyades/issues/346
Relates to https://github.com/DependencyTrack/hyades/issues/901
Relates to https://github.com/DependencyTrack/hyades/issues/907

### Additional Details

The processor defaults to ordering mode `partition` with a max concurrency matching the number of input partitions.

Ordering mode `partition` enables head-of-line blocking semantics, meaning the processor will continue retrying failing records instead of dropping them and proceeding to consume more. 

This behavior is desirable because the `dtrack.vulnerability` topic is compacted, and records are never re-published unless the respective vulnerability has changed. 

### Checklist

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -491,6 +491,19 @@ kafka.streams.transient.processing.exception.threshold.interval=PT30M
 # Refer to https://kafka.apache.org/documentation/#consumerconfigs for available options.
 # alpine.kafka.processor.<name>.consumer.<consumer.config.name>=
 
+# Required
+# Configures the Kafka processor responsible for ingesting mirrored vulnerability
+# data from the dtrack.vulnerability topic. The processor only occasionally receives
+# records, such that high concurrency is usually not justified.
+alpine.kafka.processor.vuln.mirror.max.concurrency=-1
+alpine.kafka.processor.vuln.mirror.processing.order=partition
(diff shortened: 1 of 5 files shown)


src/main/resources/application.properties
```diff
@@ -491,6 +491,19 @@ kafka.streams.transient.processing.exception.threshold.interval=PT30M
 # Refer to https://kafka.apache.org/documentation/#consumerconfigs for available options.
 # alpine.kafka.processor.<name>.consumer.<consumer.config.name>=
 
+# Required
+# Configures the Kafka processor responsible for ingesting mirrored vulnerability
```
mehab, line 495: should we also add these properties to readme since they are configurable by user?

nscuro: Good point! I'm looking into generating documentation based on contents of `application.properties` (https://github.com/DependencyTrack/hyades/issues/1154). I would consider this to be a side effect of that effort.

sahibamittal (approved):

PR #570, opened 2024-02-12
Add supplier manufacturer support

sahibamittal:
### Description

As part of backport change, add missing support for supplier and manufacturer.

### Addressed Issue

https://github.com/DependencyTrack/hyades/issues/983

### Checklist


src/main/resources/migration/changelog-v5.3.0.xml
```diff
@@ -2470,4 +2470,24 @@
...
+            <column name="SUPPLIER" type="TEXT"/>
+            <column name="AUTHORS" type="TEXT"/>
+        </createTable>
+        <addColumn tableName="COMPONENT">
+            <column name="SUPPLIER" type="TEXT"/>
+        </addColumn>
+        <addColumn tableName="PROJECT">
+            <column name="SUPPLIER" type="TEXT"/>
+            <column name="MANUFACTURER" type="TEXT"/>
+        </addColumn>
```
nscuro, line 2491: I think we should take advantage of the fact we can focus on PostgreSQL now, and use the [`JSONB`](https://www.postgresql.org/docs/current/datatype-json.html) type here, instead of `TEXT`. 

Makes it more efficient to store, query, and potentially index if we end up needing that. We already use `JSONB` for some of the `VulnerabilityPolicy` fields, so it *should just work* :D https://github.com/DependencyTrack/hyades-apiserver/blob/4e413d9bf0a77d594bcd327338ecb6c05e8a292c/src/main/resources/migration/changelog-v5.3.0.xml#L2328-L2350

nscuro (changes requested):

nscuro:
@sahibamittal Apologies, my suggestion to switch to `JSONB` caused things to break; DataNucleus can't deal with it.

Did some research on how to resolve, but it will take longer to come up with a proper solution. Documented the current status here: https://github.com/DependencyTrack/hyades/issues/1074

For now, please revert to `TEXT` again. We can switch to `JSONB` once we found a solution for https://github.com/DependencyTrack/hyades/issues/1074.

src/main/resources/migration/changelog-v5.3.0.xml
```diff
@@ -2470,4 +2470,24 @@
...
     </changeSet>
+
+    <changeSet id="v5.3.0-15" author="sahibamittal">
+        <createTable tableName="PROJECT_METADATA">
+            <column autoIncrement="true" name="ID" type="BIGINT">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="PROJECT_METADATA_PK"/>
+            </column>
+            <column name="PROJECT_ID" type="BIGINT">
+                <constraints nullable="false"/>
+            </column>
```
nscuro, line 2481: Overlooked it the first time, but we'll also need a foreign key constraint on this column.

FK constraints can be defined like so:

https://github.com/DependencyTrack/hyades-apiserver/blob/9498b4ebcbcb244e778f8b1b4f17fd9f3e774d77/src/main/resources/migration/changelog-v5.3.0.xml#L2291-L2294

nscuro (approved):

PR #573, opened 2024-02-13
Migrate all tests to use Postgres testcontainers instead of H2

nscuro:
### Description

Switches all tests to use Postgres testcontainers instead of H2.

Avoids having to deal with differences between Postgres and H2. Just adds friction considering the application will never run with H2 in production.

### Addressed Issue

Contributes to https://github.com/DependencyTrack/hyades/issues/1074

### Additional Details

Test execution times in CI only get a minor hit of 3-5min. Containers are re-used across test classes such that container restarts and database migration executions are kept to a minimum.

During this switch, multiple tests started failing due to assumptions that are true for H2 not being true for PostgreSQL. This is a win in my books, as our tests will now be closer to what is running in production.

### Checklist


sahibamittal (approved):
LGTM.

PR #580, opened 2024-02-16
Replace json string with proto in Version Distance Cel Policy

VithikaS:
### Description

PR is to replace json string with a proto in Version Distance cel policy

### Addressed Issue

This PR addresses issue https://github.com/DependencyTrack/hyades/issues/931

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/policy/cel/compat/VersionDistanceCelScriptBuilder.java
+++ b/src/main/java/org/dependencytrack/policy/cel/compat/VersionDistanceCelScriptBuilder.java
@@ -1,23 +1,54 @@
 package org.dependencytrack.policy.cel.compat;
 
+import alpine.common.logging.Logger;
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.util.JsonFormat;
+import org.apache.commons.lang3.StringUtils;
 import org.dependencytrack.model.PolicyCondition;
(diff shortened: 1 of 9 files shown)


src/main/java/org/dependencytrack/policy/cel/compat/VersionDistanceCelScriptBuilder.java
```diff
@@ -6,7 +6,19 @@ public class VersionDistanceCelScriptBuilder implements CelPolicyScriptSourceBui
     @Override
     public String apply(PolicyCondition policyCondition) {
         return """
-                component.version_distance("%s", "%s")
-                    """.formatted(policyCondition.getOperator(), CelPolicyScriptSourceBuilder.escapeQuotes(policyCondition.getValue()));
+                component.version_distance("%s", %s)
+                    """.formatted(comparator(policyCondition.getOperator()), policyCondition.getValue());
```
nscuro, line 10: To ensure compatibility with legacy conditions, we still need to expect `policyCondition.getValue()` to return JSON. The frontend still submits JSON when configuring the condition via UI, and users configuring policies via REST API will also still send JSON. The script builder's job would then be to map the provided JSON to the Protobuf struct.

nscuro (changes requested):
Looks good overall, just need an additional mapping step to ensure we're staying compatible with legacy policy conditions.

nscuro (approved):

PR #598, opened 2024-02-27
Replaced custom Vers with versatile library

sahibamittal:
### Description

While implementing the vulnerability mirroring capability, we wrote our own vers code to deal with version ranges.
Replace the custom vers handling with [versatile](https://github.com/nscuro/versatile).

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/1092

### Checklist


src/test/java/org/dependencytrack/event/kafka/streams/processor/MirrorVulnerabilityProcessorTest.java
```diff
@@ -706,7 +706,7 @@ public void testProcessVulnWithVersConstraints() throws Exception {
                             { "range": "" },
                             { "range": "<999" },
                             { "range": "vers:" },
-                            { "range": "vers:foobar/<1" },
```
nscuro, line 709: Was `foobar` causing an error?

sahibamittal: yes because Vers from versatile verifies it against Enum `VersioningScheme` and if it doesn't match it throws VersException("vers string contains an invalid versioning scheme").

nscuro: Hmmm I hadn't anticipated that, although I should have... 

So what this means is that parsing of all version ranges except [the ones explicitly supported by versatile](https://github.com/nscuro/versatile?tab=readme-ov-file#supported-versioning-schemes) will fail.

Same for the mirror-service: It won't be able to construct ranges for ecosystems for which it does not support the versioning scheme. Looking at the list of ecosystems supported by OSV, that means we're missing out on quite a lot of ranges then.

I'm not sure how realistic it is to support all possible current and future versioning schemes in versatile, that is a lot of work. I guess what we could do is fall back to `generic` whenever we encounter [this exception](https://github.com/nscuro/versatile/blob/e9f943d0297064d3a054b3dabe2002877e1f3b5f/src/main/java/io/github/nscuro/versatile/Vers.java#L77).

Any thoughts?

sahibamittal: Yes, we could fall back to generic instead of throwing exception and skipping the range. But, are we good to change the ecosystem of a version range to `generic` because it is not supported by versioningScheme? 
Like, `vers:cargo/<0.4.0` --> `vers:generic/<0.4.0`

nscuro: It's not optimal, but I don't see another solution. We basically have to decide between:

* Losing data (ignoring ranges for which we don't support the scheme)
* *Potentially* mis-interpreting the data (converting to `generic`)

The latter is the preferable one IMO. What do you think? Do you see any other options we can consider?

sahibamittal: Agreed, just one question - how do we decide the list of supported versioning schemes? Example, for OSV ecosystems like cargo or composer, in future will we be adding them? or what will be the procedure? 
Because falling back to generic will also allow test ecosystem like `foobar` to be converted to generic..

nscuro: > how do we decide the list of supported versioning schemes? Example, for OSV ecosystems like cargo or composer, in future will we be adding them?

Yes, I would assume that over time we'll add support for more ecosystems.

> Because falling back to generic will also allow test ecosystem like foobar to be converted to generic.. and there will be no use case left for VersException("vers string contains an invalid versioning scheme")

I think that is fine. It could also be a legitimate versioning scheme that we simply don't know about yet. Maybe someone comes up with one that is actually named `foobar` tomorrow :D Point being, we can't possibly foresee all the schemes, so when in doubt, I think it's better to assume `generic` instead of discarding ranges.

nscuro (approved):

PR #619, opened 2024-03-15
Migrate EPSS mirroring to hyades

sahibamittal:
### Description

EPSS mirroring is still performed by the API server. It should be performed by the mirror-service instead.

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/926

### Checklist


src/main/java/org/dependencytrack/event/kafka/streams/processor/MirrorEpssProcessor.java
```diff
@@ -0,0 +1,36 @@
...
+import io.micrometer.core.instrument.Timer;
+import org.apache.kafka.streams.processor.api.Processor;
+import org.apache.kafka.streams.processor.api.Record;
+import org.dependencytrack.model.Epss;
+import org.dependencytrack.parser.dependencytrack.EpssModelConverter;
+import org.dependencytrack.persistence.QueryManager;
+import org.dependencytrack.proto.mirror.v1.EpssItem;
+
+
+public class MirrorEpssProcessor implements Processor<String, EpssItem, Void, Void> {
```
nscuro, line 14: Consider implementing this using the new processor API, as we're already in the process of moving away from Kafka Streams :smiley: 

The code should be almost exactly the same, have a look at this PR for inspiration: https://github.com/DependencyTrack/hyades-apiserver/pull/553/files

One noteworthy difference is that you shouldn't do a broad `catch (Exception e)` anymore. This is because the processor API takes care of deciding what kind of exception can be retried, and which can not. So exceptions need to "bubble up" to it, if that makes sense. 

Also, a dedicated `Timer` is no longer necessary, as the framework will automatically track execution times of each processor.

PR #639, opened 2024-03-29
Store cluster ID in database

nscuro:
### Description

This makes the cluster ID (previously system UUID) available to other instances and services.

Instead of `Config.getInstance().getSystemUuid()`, `ClusterInfo.getClusterId()` must be used from now on.

As the cluster ID is only generated once and then never updated, it will only be loaded upon first invocation of `ClusterInfo.getClusterId()`.

Updating of the cluster ID via API or UI is prevented by marking it as read-only.

### Addressed Issue

Relates to https://github.com/DependencyTrack/hyades/issues/925

### Additional Details

N/A

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/common/ClusterInfo.java
+++ b/src/main/java/org/dependencytrack/common/ClusterInfo.java
@@ -1,0 +1,64 @@
+/*
+ * This file is part of Dependency-Track.
+ *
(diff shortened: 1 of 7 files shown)


VinodAnandan (approved):
LGTM

mehab (approved):
lgtm

PR #643, opened 2024-04-05
Ingest BOM generation timestamp

sahibamittal:
### Description

Dependency-Track currently only tracks when a BOM was uploaded, but not when the BOM was generated.

Generators can include the generation timestamp in the [metadata.timestamp](https://cyclonedx.org/docs/1.5/json/#metadata_timestamp) node of the BOM. If it is available, DT should ingest it and make it available in CEL policies.

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/1059

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/model/Bom.java
+++ b/src/main/java/org/dependencytrack/model/Bom.java
@@ -102,6 +102,10 @@ public class Bom implements Serializable {
     @NotNull
     private UUID uuid;
 
+    @Persistent
+    @Column(name = "BOM_GENERATED")
+    private Date bomGenerated;
+
     public long getId() {
         return id;
     }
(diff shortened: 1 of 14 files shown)


src/main/java/org/dependencytrack/model/Bom.java
```diff
@@ -102,6 +102,10 @@ public String getFormatLongName() {
     @NotNull
     private UUID uuid;
 
+    @Persistent
+    @Column(name = "BOM_GENERATED")
+    private Date bomGenerated;
```
nscuro, line 107: We should omit the `bom` / `BOM_` prefix here, as the field is already in the `Bom` class, and the column is in the `BOM` table.

nscuro (changes requested):
Overall looks good! Just one minor request regarding the naming of the new field and column.

nscuro (approved):

PR #654, opened 2024-04-18
Issue 947 : Add support for manual vulnerability tags

sahibamittal:
### Description

Currently, the tagging system is restricted to projects only. It is imperative to expand this functionality to include vulnerability tags, allowing for the communication of additional metadata related to vulnerabilities.

Add basic support to add/update/delete tags from a vulnerability.

### Addressed Issue

https://github.com/DependencyTrack/hyades/issues/947

### Additional Details

Requirement for automated tags is still TBD.
This PR needs corresponding frontend changes as well.

### Checklist


src/main/java/org/dependencytrack/resources/v1/VulnerabilityResource.java
```diff
@@ -618,4 +623,58 @@ public Response unassignVulnerability(@ApiParam(value = "The UUID of the vulnera
...
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(
+            value = "Updates tags for a vulnerability"
+    )
+    @ApiResponses(value = {
+            @ApiResponse(code = 401, message = "Unauthorized"),
+            @ApiResponse(code = 404, message = "The vulnerability could not be found")
+    })
+    @PermissionRequired(Permissions.Constants.VULNERABILITY_MANAGEMENT)
+    public Response updateVulnerabilityTags(Vulnerability jsonVuln) {
```
nscuro, line 639: It's a bit odd to accept an entire `Vulnerability` object, but only using two fields of it. From the outside, it will not be possible to tell which fields are used and which are not. Can we trim this down a bit?

For example:

* Change the path to `/{uuid}/tag`
* Accept a list of `String` instead of `Vulnerability` as payload

src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
```diff
@@ -887,4 +888,49 @@ public void deleteAffectedVersionAttributions(final Vulnerability vulnerability)
...
+     * Binds the two objects together in a corresponding join table.
+     *
+     * @param vulnerability a Vulnerability object
+     * @param tags    a List of Tag objects
+     */
+    @SuppressWarnings("unchecked")
+    public void bind(Vulnerability vulnerability, List<Tag> tags) {
+        final Query<Tag> query = pm.newQuery(Tag.class, "vulnerabilities.contains(:vulnerability)");
+        final List<Tag> currentVulnerabilityTags = (List<Tag>) query.execute(vulnerability);
+        pm.currentTransaction().begin();
```
nscuro, line 902: Consider using `runInTransaction` instead, as it will handle rollbacks in case something breaks between `begin` and `commit`.

src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
```diff
@@ -887,4 +888,49 @@ public void deleteAffectedVersionAttributions(final Vulnerability vulnerability)
...
+    }
+
+    /**
+     * Returns a paginated result of vulnerabilities by tag.
+     *
+     * @param tag the tag associated with the Vulnerability
+     * @return a List of vulnerabilities that contain the tag
+     */
+    public PaginatedResult getVulnerabilities(final Tag tag) {
+        final PaginatedResult result;
```
nscuro, line 925: Nit: You can omit the declaration of `result` and simply `return execute(query, params);`.

nscuro (changes requested):

nscuro (approved):
Great work! :rocket:

PR #669, opened 2024-05-01
Support multiple modes of operation for vulnerability policies

sahibamittal:
### Description

At the moment, vulnerability policies are "enforced" as soon as they are created in the system.

There may be situations however, in which policies are supposed to be deployed, but not enforced. Add a mandatory field 'mode of operation' to define the same : 

1. Disabled: Don't do anything with it, don't even consider it for evaluation.
2. Log: Evaluate policy, but don't take action, and log the outcome instead.
3. Apply: Apply policy analysis, perform db transactions and send notifications.

### Addressed Issue

https://github.com/DependencyTrack/hyades/issues/957

### Checklist

- [x] This PR introduces changes to the database model, and I have updated the [migration changelog](https://github.com/DependencyTrack/hyades-apiserver/tree/main/src/main/resources/migration) accordingly


nscuro:
@sahibamittal There are still a few `TODO`s in here, should we convert this PR to *Draft* for now?

src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
```diff
@@ -205,8 +206,16 @@ private void processScannerResult(final QueryManager qm, final Component compone
         LOGGER.debug("Identified policy matches for %d/%d vulnerabilities (scanKey: %s)"
                 .formatted(matchedPoliciesByVulnUuid.size(), syncedVulns.size(), prettyPrint(scanKey)));
 
+        // Log the matched policies with operation mode LOG
+        LOGGER.debug("List of matched vulnerability policies with mode LOG : " + matchedPoliciesByVulnUuid.values().stream().map(matchedPolicy -> matchedPolicy.getName()).toList());
```
nscuro, line 210: Let's log this as `INFO` since it's an active decision of users to enable logging for their policy.

nscuro (changes requested):

nscuro (approved):

PR #690, opened 2024-05-29
Add workflow for project cloning

sahibamittal:
### Description

Currently, when a project is cloned (via endpoint /project/clone), state on its completion is not maintained which makes it hard to determine if its cloning is finished or not.

Cloning a project is bigger task similar to bom-upload, and therefore it will be nice to have a state management for same.

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/1256

### Checklist

~~- [ ] This PR fixes a defect, and I have provided tests to verify that the fix is effective~~
- [x] This PR introduces changes to the database model, and I have updated the [migration changelog](https://github.com/DependencyTrack/hyades-apiserver/tree/main/src/main/resources/migration) accordingly
- [x] This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/hyades/tree/main/docs) accordingly


src/main/java/org/dependencytrack/tasks/CloneProjectTask.java
```diff
@@ -39,11 +44,24 @@ public void inform(final Event e) {
...
+                    workflowState.setToken(chainIdentifier);
+                    workflowState.setStartedAt(now);
+                    workflowState.setUpdatedAt(now);
+                    qm.getPersistenceManager().makePersistent(workflowState);
+                }
+                LOGGER.info("Cloning project: " + request.getProject());
                 final Project project = qm.clone(UUID.fromString(request.getProject()),
                         request.getVersion(), request.includeTags(), request.includeProperties(),
                         request.includeComponents(), request.includeServices(), request.includeAuditHistory(), request.includeACL());
+                qm.updateWorkflowStateToComplete(workflowState);
```
nscuro, line 64: We should catch exceptions thrown by `qm.clone` and transition the workflow step to `FAILED`, otherwise it will forever remain in `PENDING` state when cloning failed.

src/test/java/org/dependencytrack/tasks/CloneProjectTaskTest.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Date;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.dependencytrack.model.WorkflowStatus.COMPLETED;
+import static org.dependencytrack.model.WorkflowStep.PROJECT_CLONE;
+
+public class CloneProjectTaskTest extends PersistenceCapableTest {
+
+    @Test
+    public void testCloneProjectTask() {
```
nscuro, line 37: I think this should be covered by the clone tests in `ProjectResourceTest` already?

src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
```diff
@@ -1002,11 +1004,22 @@ public void cloneProjectTest() {
                         }
                         """.formatted(project.getUuid())));
 
-        assertThat(response.getStatus()).isEqualTo(200);
+        assertThat(response.getStatus()).isEqualTo(202);
         JsonObject json = parseJsonObject(response);
         Assert.assertNotNull(json);
         Assert.assertNotNull(json.getString("token"));
         Assert.assertTrue(UuidUtil.isValidUUID(json.getString("token")));
+        UUID uuid = UUID.fromString(json.getString("token"));
```
nscuro, line 1012: Can we have an additional test case for the scenario when cloning fails?

nscuro (changes requested):

src/test/java/org/dependencytrack/tasks/CloneProjectTaskTest.java
sahibamittal: Yes but Codacy test coverage was complaining about the `CloneProjectTask` although the code is covered by the `ProjectResourceTest`.

nscuro: Do you see the same missing coverage when using [Run with Coverage](https://www.jetbrains.com/help/idea/code-coverage.html#run_with_coverage) on the resource test?

sahibamittal: yeah locally `CloneProjectTask` is almost fully covered by `ProjectResourceTest`.

src/main/java/org/dependencytrack/tasks/CloneProjectTask.java
```diff
@@ -39,12 +44,29 @@ public void inform(final Event e) {
...
+            }
+            try {
+                LOGGER.info("Cloning project: " + request.getProject());
                 final Project project = qm.clone(UUID.fromString(request.getProject()),
                         request.getVersion(), request.includeTags(), request.includeProperties(),
                         request.includeComponents(), request.includeServices(), request.includeAuditHistory(), request.includeACL(), request.includePolicyViolations());
+                qm.updateWorkflowStateToComplete(workflowState);
                 LOGGER.info("Cloned project: " + request.getProject() + " to " + project.getUuid());
+            } catch (Exception ex) {
+                qm.updateWorkflowStateToFailed(workflowState, ex.getMessage());
```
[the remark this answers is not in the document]
nscuro: If we log the exception here, it is effectively handled and you don't need to re-throw it below. By re-throwing, it will be handled by the default uncaught exception handler, which we should avoid if possible.

src/test/java/org/dependencytrack/tasks/CloneProjectTaskTest.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @Test
+    public void testCloneProjectDoesNotExist() {
+        var uuid = UUID.randomUUID();
+        CloneProjectRequest request = new CloneProjectRequest(uuid.toString(), "1.1", false, false, false, false, false, false, false, false);
+        final var cloneProjectEvent = new CloneProjectEvent(request);
+        assertThatExceptionOfType(IllegalStateException.class)
+                .isThrownBy(() -> new CloneProjectTask().inform(cloneProjectEvent))
+                .withMessageContaining("Project with UUID " + uuid + " was supposed to be cloned, but it does not exist anymore");
```
nscuro, line 64: Following my suggestion above to not re-throw the cloning exception, this assertion will no longer work. I'd suggest to instead assert on `state.getFailureReason()`.

nscuro (approved):

PR #696, opened 2024-06-04
Include CVSS and OWASP RR vectors in notifications

VithikaS:
### Checklist


nscuro:
Thanks @VithikaS!

I think we should include these fields also for the other notifications that include vulnerability information, just to be consistent.

For the same reason, let's also include the OWASP RR vector.

VithikaS:
> Thanks @VithikaS!
> 
> I think we should include these fields also for the other notifications that include vulnerability information, just to be consistent.
> 
> For the same reason, let's also include the OWASP RR vector.



> Thanks @VithikaS!
> 
> I think we should include these fields also for the other notifications that include vulnerability information, just to be consistent.
> 
> For the same reason, let's also include the OWASP RR vector.

Added for other Notifications and owasp RR vector string

nscuro (approved):

PR #705, opened 2024-06-08
Port: Refactor BOM upload processing for better efficiency, correctness, and consistency

nscuro:
### Description

Ports https://github.com/DependencyTrack/dependency-track/pull/3357 from Dependency-Track v4.11.0

### Addressed Issue

Relates to https://github.com/DependencyTrack/hyades/issues/1190

### Additional Details

The majority of changes already existed in the Hyades codebase, since they were originally ported from Hyades to DT :)

The following changes were NOT ported:

* `IndexingInstanceLifecycleListener`: Lucene was removed in Hyades, keeping the index current is not a concern
* `L2CacheEvictingInstanceLifecycleListener`: The L2 cache has been disabled globally in Hyades, invalidating L2 cache entries is not a concern
* The `ConfigKey.BOM_PROCESSING_TASK_V2_ENABLED`: The legacy BOM processing task no longer exists in Hyades, hence making this config useless

### Checklist

- ~This PR introduces changes to the database model, and I have updated the [migration changelog](https://github.com/DependencyTrack/hyades-apiserver/tree/main/src/main/resources/migration) accordingly~
- ~This PR introduces new or alters existing behavior, and I have updated the [documentation](https://github.com/DependencyTrack/hyades/tree/main/docs) accordingly~


src/main/java/org/dependencytrack/util/InternalComponentIdentificationUtil.java
```diff
@@ -35,40 +29,8 @@ public final class InternalComponentIdentificationUtil {
...
-        final ConfigProperty internalNamesRegexProperty = qm.getConfigProperty(
-                ConfigPropertyConstants.INTERNAL_COMPONENTS_NAMES_REGEX.getGroupName(),
-                ConfigPropertyConstants.INTERNAL_COMPONENTS_NAMES_REGEX.getPropertyName()
-        );
-        if (internalNamesRegexProperty == null || StringUtils.trimToNull(internalNamesRegexProperty.getPropertyValue()) == null) {
-            return false;
-        }
-
-        return Pattern.matches(StringUtils.trimToNull(internalNamesRegexProperty.getPropertyValue()), name);
+    public static boolean isInternalComponent(final Component component) {
```
sahibamittal, line 32: Do we need this util class? We can use `InternalComponentIdentifier().isInternal` directly instead?

nscuro: Kept it mostly for backward-compatibility. But you're right, it doesn't really serve any real purpose. Let me remove it.

nscuro: Removed!

PR #709, opened 2024-06-10
Port: Apply consistent formatting to SQL query

sahibamittal:
### Description

Applies consistent formatting to SQL queries, and utilizes Java text blocks instead of string concatenation for multi-line queries.

### Addressed Issue

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/model/Finding.java
+++ b/src/main/java/org/dependencytrack/model/Finding.java
@@ -88,7 +88,7 @@ public class Finding implements Serializable {
                 ON "COMPONENT"."ID" = "COMPONENTS_VULNERABILITIES"."COMPONENT_ID"
              INNER JOIN "VULNERABILITY"
                 ON "COMPONENTS_VULNERABILITIES"."VULNERABILITY_ID" = "VULNERABILITY"."ID"
-             INNER JOIN "EPSS"
+             LEFT JOIN "EPSS"
                 ON "VULNERABILITY"."VULNID" = "EPSS"."CVE"
              INNER JOIN "FINDINGATTRIBUTION"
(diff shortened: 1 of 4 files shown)


nscuro:
The original PR modified queries that were introduced in https://github.com/DependencyTrack/dependency-track/pull/2472. Either we wait until that PR is ported, or we make sure to apply the same formatting when we port it.

src/main/java/org/dependencytrack/model/Finding.java
```diff
@@ -53,48 +53,58 @@ public class Finding implements Serializable {
...
+                 , "FINDINGATTRIBUTION"."REFERENCE_URL"
+                 , "ANALYSIS"."STATE"
+                 , "ANALYSIS"."SUPPRESSED"
+              FROM "COMPONENT"
+             INNER JOIN "COMPONENTS_VULNERABILITIES"
+                ON "COMPONENT"."ID" = "COMPONENTS_VULNERABILITIES"."COMPONENT_ID"
+             INNER JOIN "VULNERABILITY"
+                ON "COMPONENTS_VULNERABILITIES"."VULNERABILITY_ID" = "VULNERABILITY"."ID"
+             INNER JOIN "EPSS"
+                ON "VULNERABILITY"."VULNID" = "EPSS"."CVE"
```
nscuro, line 92: I think this needs to be a `LEFT JOIN` since no EPSS record may exist for a vulnerability. 

Some tests are currently failing due to missing vulnerability data, I believe that's because the tests don't create any EPSS records, hence this `INNER JOIN` yielding no records at all.

nscuro (changes requested):

nscuro (approved):

PR #712, opened 2024-06-11
Backport : support for component properties + @ValidUuid

sahibamittal:
### Description

Adds support for component properties, similar to how project properties are supported already.

On top of being manageable via REST API, component properties are also ingested upon BOM upload.

Adds support for custom annotation `@ValidUuid`.

### Addressed Issue

### Checklist


src/main/resources/migration/changelog-v5.5.0.xml
```diff
@@ -110,4 +110,27 @@
...
     </changeSet>
+
+    <changeSet id="v5.5.0-10" author="sahibamittal">
+        <createTable tableName="COMPONENT_PROPERTY">
+            <column autoIncrement="true" name="ID" type="BIGINT">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="COMPONENT_PROPERTY_PK"/>
+            </column>
+            <column name="COMPONENT_ID" type="BIGINT">
+                <constraints nullable="false"/>
+            </column>
```
nscuro, line 121: There should be a [foreign key constraint](https://docs.liquibase.com/change-types/add-foreign-key-constraint.html) for this column.

I'd vote for us to leverage `onDelete="CASCADE"` going forward. Component properties can't exist without their owning component, so when that component is deleted, the database should take care of also deleting the associated properties.

src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
```diff
@@ -266,6 +271,15 @@ public static Component convertComponent(final org.cyclonedx.model.Component cdx
...
         }
 
+        final List<ComponentProperty> properties = convertToComponentProperties(cdxComponent.getProperties());
+        if (component.getId() == 0) {
+            component.setProperties(properties);
+        } else {
+            try (var qm = new QueryManager()) {
+                qm.synchronizeComponentProperties(component, properties);
+            }
+        }
```
nscuro, line 281: Synchronization should not happen in the `ModelConverter`, we should keep persistence logic out of it as much as possible. Conversion happens in line 189 already. 

The synchronization should happen in the `BomUploadProcessingTask`, in the same DB transaction as everything else.

src/main/resources/migration/changelog-v5.5.0.xml
```diff
@@ -110,4 +110,27 @@
...
+            <column name="GROUPNAME" type="TEXT"/>
+            <column name="PROPERTYNAME" type="TEXT">
+                <constraints nullable="false"/>
+            </column>
+            <column name="PROPERTYVALUE" type="TEXT"/>
+            <column name="PROPERTYTYPE" type="TEXT">
+                <constraints nullable="false"/>
+            </column>
```
nscuro, line 129: Do we want a check constraint here, similar to what we have for `COMPONENT_CLASSIFIER_check` etc.?

nscuro (changes requested):

src/main/java/org/dependencytrack/parser/cyclonedx/util/ModelConverter.java
sahibamittal: yes, it seems `qm.synchronizeComponentProperties` is also already happening in `processComponents` in `BomUploadProcessingTask`, I'll remove this part entirely.

nscuro:
@sahibamittal Is this intended to cover https://github.com/DependencyTrack/dependency-track/pull/3590 already? There are a few more changes in that PR that are not included here.

sahibamittal:
> @sahibamittal Is this intended to cover [DependencyTrack/dependency-track#3590](https://github.com/DependencyTrack/dependency-track/pull/3590) already? There are a few more changes in that PR that are not included here.

No I didn't check that PR, do you want me to include it in this PR? Since almost all changes are here already.

nscuro:
I don't have a strong opinion on whether it should be included here. My main concern is that we don't skip it because `@ValidUuid` was already ported.

The `ConstraintViolationExceptionMapper` in https://github.com/DependencyTrack/dependency-track/pull/3590 is important for how `@ValidUuid` will behave. And the `format = "uuid"` addition for `@ApiParam` method parameters are important for how the Swagger is rendered.

sahibamittal:
> I don't have a strong opinion on whether it should be included here. My main concern is that we don't skip it because `@ValidUuid` was already ported.
> 
> The `ConstraintViolationExceptionMapper` in [DependencyTrack/dependency-track#3590](https://github.com/DependencyTrack/dependency-track/pull/3590) is important for how `@ValidUuid` will behave. And the `format = "uuid"` addition for `@ApiParam` method parameters are important for how the Swagger is rendered.

Yes, then I'll create a separate following PR for this change. I'll take that up now.

nscuro:
Awesome, thank you! Then I'll merge this one.

nscuro (approved):

PR #718, opened 2024-06-11
Customize risk score calculation

leec94:
### Description

Backend changes to customize the default risk score of vulnerabilities. Followed guidance listed here: https://github.com/DependencyTrack/dependency-track/pull/3253#issuecomment-1871503220 

### Addressed Issue

Overall issue https://github.com/DependencyTrack/dependency-track/issues/2824 

### Additional Details

Corresponding frontend changes here: https://github.com/DependencyTrack/hyades-frontend/pull/54

### Checklist


leec94:
hi @nscuro do you have any advice on adding tests for this?

i added for custom risk scores, making MetricsTest extend PersistenceCapableTest so the querymanager could be used, but having issues. here's the error i'm getting:

```
java.lang.reflect.InaccessibleObjectException: Unable to make field final java.util.Map java.util.Collections$UnmodifiableMap.m accessible: module java.base does not "opens java.util" to unnamed module @41a810ee
```

src/test/java/org/dependencytrack/metrics/MetricsTest.java
```diff
@@ -18,10 +18,23 @@
...
+import static org.dependencytrack.model.ConfigPropertyConstants.CUSTOM_RISK_SCORE_CRITICAL;
+import static org.dependencytrack.model.ConfigPropertyConstants.CUSTOM_RISK_SCORE_HIGH;
+import static org.dependencytrack.model.ConfigPropertyConstants.CUSTOM_RISK_SCORE_MEDIUM;
+import static org.dependencytrack.model.ConfigPropertyConstants.CUSTOM_RISK_SCORE_LOW;
+import static org.dependencytrack.model.ConfigPropertyConstants.CUSTOM_RISK_SCORE_UNASSIGNED;
+
+import org.dependencytrack.PersistenceCapableTest;
+
+public class MetricsTest extends PersistenceCapableTest{
+    protected QueryManager qm;
```
nscuro, line 37: Removing this should resolve the current error in CI. `qm` should be inherited from `PersistenceCapableTest` and be accessible to you in your test method.

nscuro:
@leec94 Can you double-check that you're using JDK 21 locally?

The error I see in the CI build logs say:

```
Error:  testCustomRiskScores(org.dependencytrack.metrics.MetricsTest)  Time elapsed: 0.357 s  <<< ERROR!
java.lang.NullPointerException: Cannot invoke "org.dependencytrack.persistence.QueryManager.createConfigProperty(String, String, String, alpine.model.IConfigProperty$PropertyType, String)" because "this.qm" is null
	at org.dependencytrack.metrics.MetricsTest.testCustomRiskScores(MetricsTest.java:50)
```

I added a comment on the code area which I think should resolve this.

src/main/java/org/dependencytrack/metrics/Metrics.java
```diff
@@ -36,11 +46,30 @@
...
+            int risk_score_high_val = Integer.valueOf(risk_score_high_property.getPropertyValue());
+            int risk_score_medium_val = Integer.valueOf(risk_score_medium_property.getPropertyValue());
+            int risk_score_low_val = Integer.valueOf(risk_score_low_property.getPropertyValue());
+            int risk_score_unassigned_val = Integer.valueOf(risk_score_unassigned_property.getPropertyValue());
+
+            return (double) ((critical * risk_score_critical_val) + (high * risk_score_high_val) + (medium * risk_score_medium_val) + (low * risk_score_low_val) + (unassigned * risk_score_unassigned_val));
+
+        } catch (Exception e) {
+            return (double) ((critical * 10) + (high * 5) + (medium * 3) + (low * 1) + (unassigned * 5)); 
+        }
```
nscuro, line 72: Note, I think this method isn't used anymore.

We had to move metrics calculation to stored procedures on the database, since moving all this data between database and API server was too resource-intensive.

This corresponding SQL for this method now lives here: https://github.com/DependencyTrack/hyades-apiserver/blob/df57c53cd2bb91f6bc8e642d6a1bea353153da9e/src/main/resources/migration/procedures/function_calc-risk-score.sql#L1-L19

It is invoked during component metrics calculation, here:

https://github.com/DependencyTrack/hyades-apiserver/blob/df57c53cd2bb91f6bc8e642d6a1bea353153da9e/src/main/resources/migration/procedures/procedure_update-component-metrics.sql#L123

In theory, you can get your desired outcome with something like this (not tested):

```sql
WITH "CUSTOM_SCORES" AS (
  SELECT "PROPERTYVALUE"::INT AS "value"
       , "PROPERTYNAME" AS "name"
    FROM "CONFIGPROPERTY"
   WHERE "PROPERTYGROUP" = 'risk-score'
     AND "PROPERTYTYPE" = 'INTEGER'
)
SELECT (
  ("critical" * (SELECT "value" FROM "CUSTOM_SCORES" WHERE "name" = 'riskscore.critical'))
  -- TODO: Repeat for the other scores
  + ("high" * 5) 
  + ("medium" * 3)
  + ("low" * 1)
  + ("unassigned" * 5)
)::NUMERIC;
```

This uses a [CTE](https://www.postgresql.org/docs/current/queries-with.html) to get the relevant config values.

nscuro: You can still test this in mostly-Java. The [metrics task tests](https://github.com/DependencyTrack/hyades-apiserver/blob/main/src/test/java/org/dependencytrack/tasks/metrics/ComponentMetricsUpdateTaskTest.java) all call the stored procedures behind the scenes. So you can do the test setup, and assertions of your desired outcome as usual.

leec94: i removed the qm line and now the metricstests pass, thanks. though maybe i should revert it all since that function isn't used anymore. 

i added the sql, but i didn't know if that quotes are still needed, and if it should be in the create function statement or a separate one. is there a way to test just the sql?

src/test/java/org/dependencytrack/metrics/MetricsTest.java
leec94: removed, thanks

leec94:
> @leec94 Can you double-check that you're using JDK 21 locally?
> 
> The error I see in the CI build logs say:
> 
> ```
> Error:  testCustomRiskScores(org.dependencytrack.metrics.MetricsTest)  Time elapsed: 0.357 s  <<< ERROR!
> java.lang.NullPointerException: Cannot invoke "org.dependencytrack.persistence.QueryManager.createConfigProperty(String, String, String, alpine.model.IConfigProperty$PropertyType, String)" because "this.qm" is null
> 	at org.dependencytrack.metrics.MetricsTest.testCustomRiskScores(MetricsTest.java:50)
> ```
> 
> I added a comment on the code area which I think should resolve this.

vscode says i'm running JavaSE-21, but the errors are still showing, and inconsistent with the CI pipeline errors. it seems like something is wrong with my local environment but can't seem to see what it is

nscuro:
> vscode says i'm running JavaSE-21, but the errors are still showing, and inconsistent with the CI pipeline errors. it seems like something is wrong with my local environment but can't seem to see what it is

Is there more output when this error occurs? If there's a stacktrace or something I may be able to help pinpointing the issue. Alternatively, if you can share a few more details about your setup (i.e. which VSCode extension you're using) I can try to reproduce.

src/main/java/org/dependencytrack/metrics/Metrics.java
nscuro: The SQL in your most recent commit looks like it should work. There are test failures now because the new `ConfigProperty`s are not created in those tests. Once you create the properties for them, it should work!

> is there a way to test just the sql?

You can always create / replace the SQL function in a database and manually invoke it. But IMO the easiest way to test it is to use JUnit, as done in the metrics task tests I linked above. The functions are automatically deployed when the test start.

leec94:
i managed to run my frontend changes https://github.com/DependencyTrack/hyades-frontend/pull/54 with the latest hyades and this branch of hyades-apiserver, and was able to run DT and view the risk score settings page

image

and changing the score logs an event
image

src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
```diff
@@ -94,7 +94,13 @@ public enum ConfigPropertyConstants {
...
+    CUSTOM_RISK_SCORE_HISTORY_ENABLED("risk-score", "riskscore.history.enabled", "true", PropertyType.BOOLEAN, "Flag to re-calculate risk score history", ConfigPropertyAccessMode.READ_WRITE),
+    CUSTOM_RISK_SCORE_CRITICAL("risk-score", "riskscore.critical", "10", PropertyType.INTEGER, "Critical severity vulnerability weight (between 1-10)", ConfigPropertyAccessMode.READ_WRITE),
+    CUSTOM_RISK_SCORE_HIGH("risk-score", "riskscore.high", "5", PropertyType.INTEGER, "High severity vulnerability weight (between 1-10)", ConfigPropertyAccessMode.READ_WRITE),
+    CUSTOM_RISK_SCORE_MEDIUM("risk-score", "riskscore.medium", "3", PropertyType.INTEGER, "Medium severity vulnerability weight (between 1-10)", ConfigPropertyAccessMode.READ_WRITE),
+    CUSTOM_RISK_SCORE_LOW("risk-score", "riskscore.low", "1", PropertyType.INTEGER, "Low severity vulnerability weight (between 1-10)", ConfigPropertyAccessMode.READ_WRITE),
+    CUSTOM_RISK_SCORE_UNASSIGNED("risk-score", "riskscore.unassigned", "5", PropertyType.INTEGER, "Unassigned severity vulnerability weight (between 1-10)", ConfigPropertyAccessMode.READ_WRITE);
```
nscuro, line 103: Consider removing the redundant `riskscore` in the property name.

Also, should we make the naming a bit more explicit to express that these are *weights*?

e.g. group name: `risk-score`, property name: `weight.critical`

src/main/java/org/dependencytrack/resources/v1/AbstractConfigPropertyResource.java
```diff
@@ -75,6 +80,36 @@ private Response updatePropertyValueInternal(IConfigProperty json, IConfigProper
...
+                }
+                if (ConfigPropertyConstants.CUSTOM_RISK_SCORE_LOW.getPropertyName().equals(json.getPropertyName())){
+                    super.logSecurityEvent(LOGGER, SecurityMarkers.SECURITY_AUDIT, "Risk score \"" + json.getPropertyName() + "\" changed to value: " + propertyValue + " by user " + super.getPrincipal().getName());
+                }
+                if (ConfigPropertyConstants.CUSTOM_RISK_SCORE_UNASSIGNED.getPropertyName().equals(json.getPropertyName()) && (propertyValue < 1 || propertyValue > 10)){
+                    return Response.status(Response.Status.BAD_REQUEST).entity("Risk score \""+json.getPropertyName()+"\" must be between 1 and 10. An invalud value of " + propertyValue + " was provided.").build();
+                }
+                if (ConfigPropertyConstants.CUSTOM_RISK_SCORE_UNASSIGNED.getPropertyName().equals(json.getPropertyName()) && (propertyValue < 1 || propertyValue > 10)){
+                    super.logSecurityEvent(LOGGER, SecurityMarkers.SECURITY_AUDIT, "Risk score \"" + json.getPropertyName() + "\" changed to value: " + propertyValue + " by user " + super.getPrincipal().getName());
+                }
```
nscuro, line 112: Can we generalize this such that we don't need an `if` condition for each property? The checks are the same for all weight properties.

src/test/java/org/dependencytrack/event/kafka/processor/ProcessedVulnerabilityScanResultProcessorTest.java
```diff
@@ -180,6 +188,13 @@ public void testProcessWithResultWithoutScannerResults() throws Exception {
         project.setName("acme-app");
         qm.persist(project);
 
+        qm.createConfigProperty(CUSTOM_RISK_SCORE_CRITICAL.getGroupName(), CUSTOM_RISK_SCORE_CRITICAL.getPropertyName(), "10", IConfigProperty.PropertyType.INTEGER, null);
+        qm.createConfigProperty(CUSTOM_RISK_SCORE_HIGH.getGroupName(), CUSTOM_RISK_SCORE_HIGH.getPropertyName(), "5", IConfigProperty.PropertyType.INTEGER, null);
+        qm.createConfigProperty(CUSTOM_RISK_SCORE_MEDIUM.getGroupName(), CUSTOM_RISK_SCORE_MEDIUM.getPropertyName(), "3", IConfigProperty.PropertyType.INTEGER, null);
+        qm.createConfigProperty(CUSTOM_RISK_SCORE_LOW.getGroupName(), CUSTOM_RISK_SCORE_LOW.getPropertyName(), "1", IConfigProperty.PropertyType.INTEGER, null);
+        qm.createConfigProperty(CUSTOM_RISK_SCORE_UNASSIGNED.getGroupName(), CUSTOM_RISK_SCORE_UNASSIGNED.getPropertyName(), "5", IConfigProperty.PropertyType.INTEGER, null);
```
nscuro, line 195: This test does not need these properties. Sadly it is flaky every now and then, which is why it failed in a previous CI run.

src/test/java/org/dependencytrack/tasks/metrics/PortfolioMetricsUpdateTaskTest.java
```diff
@@ -62,6 +69,12 @@ public static void tearDownClass() {
 
     @Test
     public void testUpdateMetricsEmpty() {
+        qm.createConfigProperty(CUSTOM_RISK_SCORE_CRITICAL.getGroupName(), CUSTOM_RISK_SCORE_CRITICAL.getPropertyName(), "10", IConfigProperty.PropertyType.INTEGER, null);
+        qm.createConfigProperty(CUSTOM_RISK_SCORE_HIGH.getGroupName(), CUSTOM_RISK_SCORE_HIGH.getPropertyName(), "5", IConfigProperty.PropertyType.INTEGER, null);
+        qm.createConfigProperty(CUSTOM_RISK_SCORE_MEDIUM.getGroupName(), CUSTOM_RISK_SCORE_MEDIUM.getPropertyName(), "3", IConfigProperty.PropertyType.INTEGER, null);
+        qm.createConfigProperty(CUSTOM_RISK_SCORE_LOW.getGroupName(), CUSTOM_RISK_SCORE_LOW.getPropertyName(), "1", IConfigProperty.PropertyType.INTEGER, null);
+        qm.createConfigProperty(CUSTOM_RISK_SCORE_UNASSIGNED.getGroupName(), CUSTOM_RISK_SCORE_UNASSIGNED.getPropertyName(), "5", IConfigProperty.PropertyType.INTEGER, null);
```
nscuro, line 76: Consider replacing literals (e.g. `"10"`, `IConfigProperty.PropertyType.INTEGER`) with the properties' getters instead. That ensures that changes made to `ConfigPropertyConstants` also apply to tests.

So, instead of:

```java
qm.createConfigProperty(CUSTOM_RISK_SCORE_CRITICAL.getGroupName(), CUSTOM_RISK_SCORE_CRITICAL.getPropertyName(), "10", IConfigProperty.PropertyType.INTEGER, null);
```

Consider doing:

```
qm.createConfigProperty(
    CUSTOM_RISK_SCORE_CRITICAL.getGroupName(), 
    CUSTOM_RISK_SCORE_CRITICAL.getPropertyName(), 
    CUSTOM_RISK_SCORE_CRITICAL.getDefaultPropertyValue(), 
    CUSTOM_RISK_SCORE_CRITICAL.getPropertyType(), 
    CUSTOM_RISK_SCORE_CRITICAL.getDescription()
);
```

If this takes too much visual space, consider moving it to a method in `AbstractMetricsUpdateTaskTest`, so all `*MetricsUpdateTaskTest` classes can make use of it.

nscuro (changes requested):

src/main/java/org/dependencytrack/metrics/Metrics.java
leec94: moved the logic to sql and updated ComponentMetricsUpdateTaskTest

src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
leec94: updated property names and in frontend too

src/main/java/org/dependencytrack/resources/v1/AbstractConfigPropertyResource.java
leec94: reduced number of if statements, tested logic and its still works the same

src/test/java/org/dependencytrack/tasks/metrics/PortfolioMetricsUpdateTaskTest.java
leec94: updated to use the property getters and used a helper function to create the properties

src/test/java/org/dependencytrack/event/kafka/processor/ProcessedVulnerabilityScanResultProcessorTest.java
leec94: removed

nscuro:
Thanks @leec94, looks good!

Do you think we can make the diff coverage check happy still?

We probably want at least one test that covers the new conditions in `AbstractConfigPropertyResource`, that can lead to a `BAD_REQUEST` response. You can find an example test for that here:

https://github.com/DependencyTrack/hyades-apiserver/blob/a9abb9f55fe51b3fef2fcc33c81754ccabbbcfa7/src/test/java/org/dependencytrack/resources/v1/ConfigPropertyResourceTest.java#L213-L235

Note that we can't test logs, so you likely can't cover everything there.

leec94:
> Thanks @leec94, looks good!
> 
> Do you think we can make the diff coverage check happy still?
> 
> We probably want at least one test that covers the new conditions in `AbstractConfigPropertyResource`, that can lead to a `BAD_REQUEST` response. You can find an example test for that here:
> 
> https://github.com/DependencyTrack/hyades-apiserver/blob/a9abb9f55fe51b3fef2fcc33c81754ccabbbcfa7/src/test/java/org/dependencytrack/resources/v1/ConfigPropertyResourceTest.java#L213-L235
> 
> Note that we can't test logs, so you likely can't cover everything there.

you've been so helpful guiding me though this, i really appreciate it! i added a few tests, but we'll see if the coverage test wants me to add more.

nscuro (approved):
Great work @leec94!

PR #742, opened 2024-06-17
Port: Webhook alert token and new user alerts

sahibamittal:
### Description

This PR adds two new notification types (user_created and user_deleted) and the possibility of setting a token on the webhook alert.

### Addressed Issue

### Checklist


src/main/java/org/dependencytrack/event/kafka/KafkaEventConverter.java
```diff
@@ -245,6 +248,11 @@ private static String extractEventKey(final Notification notification) throws In
                 final var subject = notification.getSubject().unpack(VexConsumedOrProcessedSubject.class);
                 yield requireNonEmpty(subject.getProject().getUuid());
             }
+            case GROUP_USER_CREATED, GROUP_USER_DELETED -> {
+                requireSubjectOfTypeAnyOf(notification, List.of(UserPrincipalSubject.class));
+                final var subject = notification.getSubject().unpack(UserPrincipalSubject.class);
+                yield requireNonEmpty(subject.getUsername());
```
nscuro, line 254: Good choice!

src/main/proto/org/dependencytrack/notification/v1/notification.proto
```diff
@@ -216,6 +218,11 @@ message ProjectVulnAnalysisCompleteSubject {
   string token = 4;
 }
 
+message UserPrincipalSubject {
```
nscuro, line 221: Nit: Consider just calling it `UserSubject`. The `Principal` part is more of a [Java-specific thing](https://www.baeldung.com/security-context-basics) that IMO should not "bleed" into our public APIs.

src/main/java/org/dependencytrack/parser/dependencytrack/NotificationModelConverter.java
```diff
@@ -182,6 +190,8 @@ private static Optional<com.google.protobuf.Any> convert(final Object subject) {
             return Optional.of(Any.pack(convert(pvi)));
         } else if (subject instanceof final org.dependencytrack.model.Project p) {
             return Optional.of(Any.pack(convert(p)));
+        } else if (subject instanceof final UserPrincipalSubject p) {
+            return Optional.of(Any.pack(convert(p)));
```
nscuro, line 194: Looks like we don't need any conversion here since the subject is already a Proto struct.

```suggestion
        } else if (subject instanceof final UserPrincipalSubject p) {
            return Optional.of(Any.pack(p));
```

nscuro (changes requested):

PR #748, opened 2024-06-18
Port : Truncate component property value

sahibamittal:
### Description

Truncates ComponentProperty values.

This is both to prevent the database's length constraint from being violated, and to avoid storing unnecessary long values. Some tools may misuse properties, and in those cases having the complete value is usually not of interest to users.

### Addressed Issue

### Checklist

### DIFF
--- a/src/test/java/org/dependencytrack/resources/v1/FindingResourceTest.java
+++ b/src/test/java/org/dependencytrack/resources/v1/FindingResourceTest.java
@@ -36,6 +36,8 @@ import org.dependencytrack.model.Severity;
 import org.dependencytrack.model.Vulnerability;
 import org.dependencytrack.model.WorkflowStep;
 import org.glassfish.jersey.server.ResourceConfig;
+import org.json.JSONArray;
+import org.json.JSONObject;
 import org.junit.Assert;
 import org.junit.ClassRule;
 import org.junit.Test;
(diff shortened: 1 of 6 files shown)


nscuro:
The limit should also apply to the database schema:

https://github.com/DependencyTrack/hyades-apiserver/blob/e0f083e3d2ffef0c30cba22fba6b36e4148f2aa5/src/main/resources/migration/changelog-v5.5.0.xml#L115-L134

On the same note, could you also add the limits for the other fields so they're in sync with what's defined on the Java model?

sahibamittal:
> The limit should also apply to the database schema:
> 
> https://github.com/DependencyTrack/hyades-apiserver/blob/e0f083e3d2ffef0c30cba22fba6b36e4148f2aa5/src/main/resources/migration/changelog-v5.5.0.xml#L115-L134
> 
> On the same note, could you also add the limits for the other fields so they're in sync with what's defined on the Java model?

Sure, though fields other than `propertyValue` don't have length constraints in java model, do you want me add them like in upstream?

nscuro:
Argh, apologies I was looking at the upstream code indeed. I think it makes sense to add them for consistency, WDYT?

nscuro (dismissed):

nscuro:
Seems the new SARIF test is a bit flaky:

```
 Error:  Failures: 
Error:    FindingResourceTest.getSARIFFindingsByProjectTest:695 JSON documents are different:
Different value found in node "runs[0].results[1].level", expected: <"note"> but was: <"error">.
Different value found in node "runs[0].results[1].message.text", expected: <"A description-with-hyphens-(and parentheses)"> but was: <"Yet another description but with surrounding whitespaces">.
Different value found in node "runs[0].results[1].properties.cweId", expected: <"23"> but was: <"46">.
Different value found in node "runs[0].results[1].properties.recommendation", expected: <"Recommendation with whitespaces"> but was: <"">.
Different value found in node "runs[0].results[1].properties.severityRank", expected: <"3"> but was: <"1">.
Different value found in node "runs[0].results[1].ruleId", expected: <"Vuln-3"> but was: <"Vuln-2">.
Different value found in node "runs[0].results[2].level", expected: <"error"> but was: <"note">.
Different value found in node "runs[0].results[2].message.text", expected: <"Yet another description but with surrounding whitespaces"> but was: <"A description-with-hyphens-(and parentheses)">.
Different value found in node "runs[0].results[2].properties.cweId", expected: <"46"> but was: <"23">.
Different value found in node "runs[0].results[2].properties.recommendation", expected: <""> but was: <"Recommendation with whitespaces">.
Different value found in node "runs[0].results[2].properties.severityRank", expected: <"1"> but was: <"3">.
Different value found in node "runs[0].results[2].ruleId", expected: <"Vuln-2"> but was: <"Vuln-3">.
```

Do you see any way we can stabilize it?

src/test/java/org/dependencytrack/resources/v1/FindingResourceTest.java
```diff
@@ -688,174 +690,135 @@ public void getSARIFFindingsByProjectTest() {
         Assert.assertEquals(200, response.getStatus(), 0);
         Assert.assertEquals(MEDIA_TYPE_SARIF_JSON, response.getHeaderString(HttpHeaders.CONTENT_TYPE));
         final String jsonResponse = getPlainTextBody(response);
+        JSONArray resultArray = new JSONObject(jsonResponse).getJSONArray("runs").getJSONObject(0).getJSONArray("results");
 
         assertThatJson(jsonResponse)
                 .withMatcher("version", equalTo(new About().getVersion()))
-                .withMatcher("fullName", equalTo("OWASP Dependency-Track - " + new About().getVersion()))
```
nscuro, line 694: You can achieve the same outcome without the need to parse intermediary objects with:

https://github.com/DependencyTrack/hyades-apiserver/blob/a48b7973514b9c07bf6a84de24651fac8e395dce/src/test/java/org/dependencytrack/model/mapping/PolicyProtoMapperTest.java#L79

nscuro (approved):

PR #751, opened 2024-06-20
Fix missing argument list for `DROP FUNCTION` migrations

nscuro:
### Description

Fixes missing argument list for `DROP FUNCTION` migrations.

`CALC_SEVERITY` can exist with both 3, and 4 arguments, depending on when the system was first deployed. `CALC_SEVERITY` was updated to take an additional `"severity_override" VARCHAR` argument in https://github.com/DependencyTrack/hyades-apiserver/commit/6845ea1b7c11f163d9d712fdeaac9593fc903469#diff-8a05bac09ccef76206adca46106e47020121bc13254146fc287fd1279ddfad8a, but the previous version of the function was not dropped back then.

This adds argument lists to all functions being dropped, to avoid ambiguity.

### Addressed Issue

N/A

### Additional Details

N/A

### Checklist

### DIFF
--- a/src/main/resources/migration/changelog-v5.5.0.xml
+++ b/src/main/resources/migration/changelog-v5.5.0.xml
@@ -145,9 +145,10 @@
     <changeSet id="v5.5.0-11" author="nscuro">
         <customChange class="org.dependencytrack.persistence.migration.change.v550.ComputeSeveritiesChange"/>
         <sql splitStatements="true">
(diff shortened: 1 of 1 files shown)


VithikaS (approved):

nscuro:
The failing test is `getSARIFFindingsByProjectTest` which was found to be flaky, and should be fixed in #748. Merging this anyway.

PR #785, opened 2024-07-18
Migrate to Jakarta EE and OpenAPI v3

mehab:
### Description
This PR backports changes from https://github.com/DependencyTrack/dependency-track/pull/3726 and https://github.com/DependencyTrack/dependency-track/pull/3730 into hyades-apiserver

### Checklist


sahibamittal (approved):
lgtm 👍🏼

pom.xml
```diff
@@ -691,12 +698,13 @@
...
             <plugin>
-                <groupId>org.eclipse.jetty</groupId>
-                <artifactId>jetty-maven-plugin</artifactId>
+                <groupId>org.eclipse.jetty.ee10</groupId>
+                <artifactId>jetty-ee10-maven-plugin</artifactId>
                 <version>${plugin.jetty.version}</version>
                 <configuration>
                     <webApp>
                         <contextPath>/</contextPath>
+                        <overrideDescriptor>src/test/webapp/WEB-INF/h2-console-web.xml</overrideDescriptor>
```
nscuro, line 707: The [file](https://github.com/DependencyTrack/dependency-track/pull/3730/files#diff-eca470f69b8d1376acb646175eee995ad2b84f99655a08c7867b896ec6f26c87) does not exist in the PR. And because we removed H2 entirely, I think you can simply omit this line.

src/main/resources/META-INF/persistence.xml
```diff
@@ -17,48 +17,49 @@
   ~ SPDX-License-Identifier: Apache-2.0
   ~ Copyright (c) OWASP Foundation. All Rights Reserved.
   -->
-<persistence xmlns="http://xmlns.jcp.org/xml/ns/persistence"
+<persistence xmlns="https://jakarta.ee/xml/ns/persistence"
```
nscuro, line 20: `xmlns` should remain at `http://xmlns.jcp.org/xml/ns/persistence`. Essentially, DataNucleus (which is using this file) has not yet moved to Jakarta itself. So this should remain at its pre-Jakarta value. I ran into this too, during the migration work in v4.12. :sweat_smile:

nscuro, line 20: `xmlns` should remain at `http://xmlns.jcp.org/xml/ns/persistence`. Essentially, DataNucleus (which is using this file) has not yet moved to Jakarta itself. So this should remain at its pre-Jakarta value. I ran into this too, during the migration work in v4.12. :sweat_smile:

src/main/java/org/dependencytrack/persistence/PolicyQueryManager.java
```diff
@@ -398,7 +398,7 @@ public PaginatedResult getPolicyViolations(boolean includeSuppressed) {
         }
         final PaginatedResult result = execute(query);
         for (final PolicyViolation violation: result.getList(PolicyViolation.class)) {
-            violation.getPolicyCondition().getPolicy(); // force policy to ne included since its not the default
+            violation.getPolicyCondition().getPolicy(); // force policy to be included since its not the default
```
nscuro, line 401: :nerd_face:

nscuro (changes requested):
Thanks for taking on this daunting work @mehab, really much appreciated! :pray: 

Found a few minor wrinkles, but overall looks good.

src/main/resources/META-INF/persistence.xml
nscuro: Whoops, duplicate comment.

nscuro (approved):

PR #789, opened 2024-07-19
Re-calculate historical risk score

leec94:
### Description

Related to https://github.com/DependencyTrack/dependency-track/issues/2824 , this issue is to allow the custom weights of risk scores to be re-calculated. 

### Addressed Issue

Related to https://github.com/DependencyTrack/dependency-track/issues/2824 

Front end issue to go along with this is https://github.com/DependencyTrack/hyades-frontend/pull/94 

### Additional Details
Adds a new endpoint to tell the database to recalculate with `/riskscore/refresh`. If there's a better approach for this, I'm all ears! 

This PR is draft, I have a few questions / open ended items: 
- I'm not super experienced with SQL so I'm confused about how to define the columns to make sure the risk scores are recalculated throughout the `DEPENDENCYMETRICS` table. 
- I need to only run this recalculation if `weight.history.enabled` is True, don't run this when false. Where would this logic be done?
- Tests haven't been added yet

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/event/HistoricalRiskScoreUpdateEvent.java
+++ b/src/main/java/org/dependencytrack/event/HistoricalRiskScoreUpdateEvent.java
@@ -1,0 +1,46 @@
+/*
+ * This file is part of Dependency-Track.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) OWASP Foundation. All Rights Reserved.
+ */
+package org.dependencytrack.event;
+
+import alpine.event.framework.Event;
+import alpine.event.framework.SingletonCapableEvent;
+
+/**
+ * Defines an {@link Event} used to trigger historical risk score metrics updates.
+ *
+ * @since 5.0.0
+ */
+public class HistoricalRiskScoreUpdateEvent extends SingletonCapableEvent {
+
+    private final boolean weightHistoryEnabled;
+
+    public HistoricalRiskScoreUpdateEvent() {
+        this(true);
+    }
+
+    public HistoricalRiskScoreUpdateEvent(final boolean weightHistoryEnabled) {
+        this.setSingleton(true);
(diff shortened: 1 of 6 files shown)


nscuro:
> I have pgadmin set up and connected to the hyades database, but i'm unable to run a query of the sql as it is, or even just the "update" part.

Can you share some more details on the issues you're encountering? What are you trying, and is pgadmin showing any errors?

> i'm not sure how to define the columns to make sure the risk scores are recalculated throughout the DEPENDENCYMETRICS table.

I'll try to provide some guidance below.

---

In general I think the whole thing becomes easier if you don't use a stored procedure for this. [The SQL you already have](https://github.com/DependencyTrack/hyades-apiserver/pull/789/files#diff-b47c6c3133fcc92d16b346311080fb897cc643af0e63a5e1691c30c9d0b5bf1f):

```sql
UPDATE "DEPENDENCYMETRICS" SET "v_risk_score" = "CALC_RISK_SCORE"("v_critical", "v_high", "v_medium", "v_low", "v_unassigned");
```

can be rewritten to the following:

```sql
UPDATE "DEPENDENCYMETRICS"
   SET "RISK_SCORE" = "CALC_RISK_SCORE"("CRITICAL", "HIGH", "MEDIUM", "LOW", "UNASSIGNED_SEVERITY")
```

There is no need to declare any input or output variables, the RDBMS will execute the above as-is, and update the risk score of all records in the `DEPENDENCYMETRICS` table accordingly.

As for how to execute the query, the main way we interact with the database now is via [JDBI](https://jdbi.org/). It should also "just work" in pgadmin, provided you have data in your `DEPENDENCYMETRICS` table.

There are multiple ways to use JDBI, but the [fluent API](https://jdbi.org/#_fluent_api) is probably the easiest. You can find an example here:

https://github.com/DependencyTrack/hyades-apiserver/blob/f36392b16e5b7eca0ac42c52d377036b4878dc57/src/test/java/org/dependencytrack/resources/v1/AnalysisResourceTest.java#L808-L820 

*(Ignore the query in the example above, just want to showcase how to use `createUpdate` with JDBI)*

---

I personally find that working on something like this is easiest in a test-driven way:

1. Create a test that sets up your input data. In your case, you need a project, a component in that project, and an existing `DependencyMetrics` record for the component.
2. Execute your new code on the dataset above.
3. Assert that the dataset has changed according to your expectation. In your case, you will want to ensure that the `riskScore` values were updated correctly.

You can take inspiration from the [existing metrics update tests](https://github.com/DependencyTrack/hyades-apiserver/blob/f36392b16e5b7eca0ac42c52d377036b4878dc57/src/test/java/org/dependencytrack/tasks/metrics/ComponentMetricsUpdateTaskTest.java#L141-L222).

---

> I need to only run this recalculation if weight.history.enabled is True, don't run this when false. Where would this logic be done?

Add this as a check to the new `HistoricalRiskScoreUpdateTask`. Consider logging a warning when the task is executed, but the feature itself is disabled.

--- 

> Tests haven't been added yet

See above. Getting tests set up will help you in developing the feature. Let me know if you need more help with that.

PR #805, opened 2024-07-27
Introduce plugin system to deal with provider config and lifecycle

nscuro:
### Description

Introduce a plugin system to better deal with provider configuration and lifecycle.

We already have a handful of features with multiple "implementations". Users can choose their preferred implementation, depending on their specific needs and requirements:

* Vulnerability policy bundles can be fetched from S3, or generic file servers.
* Uploaded BOMs can be stored on local disk, in the database, or S3.

As we continue to build out Dependency-Track as a platform, cases like the above will increase. For example, the following use-cases may benefit from a plugin mechanism:

* Support for multiple BOM formats (SPDX, Syft, ...).
* ...

We need a unified mechanism to deal with such extensibility demands.

> [!NOTE]
> This implementation does not yet support loading of plugins from JARs on the filesystem. It also does not yet enable users to build custom plugins. To do that, we need to decouple plugin interfaces into a separate Maven module, and publish that to Maven Central. This in turn requires the API server to switch to a multi-module Maven project model.

### Addressed Issue

N/A

### Additional Details

The implementation is based on, or heavily inspired by other plugin systems that I found in-the-wild. Particularly the following:

* [Kafka Connect plugins](https://github.com/apache/kafka/tree/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/isolation)
* [Keycloak providers](https://github.com/keycloak/keycloak/tree/main/server-spi/src/main/java/org/keycloak/provider)
* [Nzyme plugins](https://github.com/nzymedefense/nzyme-plugin/tree/master/src/main/java/app/nzyme/plugin)

### Checklist


hoggmania:
Look's good to go to me.

sahibamittal (approved):
Great work 👍🏼

PR #836, opened 2024-08-08
Port : Search component by group

sahibamittal:
### Description

On the `Components` tab, search components also by group. Today it only searches by name but some components use generic names and differentiate themselves by group. For example, `@angular/corehas` group `https://github.com/angular` and name `core`, so searching for `angular` I don't get all the Angular packages on the results.

### Addressed Issue

Backport : https://github.com/DependencyTrack/hyades/issues/1358

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
@@ -240,8 +240,8 @@ final class ComponentQueryManager extends QueryManager implements IQueryManager 
         queryParams.put("projectId", project.getId());
         if (filter != null) {
             queryString += """
-                    AND LOWER("A0"."NAME") LIKE ('%' || LOWER(:nameFilter) || '%')
-                    OR LOWER("A0"."GROUP") LIKE ('%' || LOWER(:nameFilter) || '%')
+                    AND (LOWER("A0"."NAME") LIKE ('%' || LOWER(:nameFilter) || '%')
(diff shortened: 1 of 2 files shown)


src/main/java/org/dependencytrack/persistence/ComponentQueryManager.java
```diff
@@ -241,6 +241,7 @@ public PaginatedResult getComponents(final Project project, final boolean includ
         if (filter != null) {
             queryString += """
                     AND LOWER("A0"."NAME") LIKE ('%' || LOWER(:nameFilter) || '%')
+                    OR LOWER("A0"."GROUP") LIKE ('%' || LOWER(:nameFilter) || '%')
```
nscuro, line 244: The filters should be enclosed in parentheses:

```suggestion
                    AND (LOWER("A0"."NAME") LIKE ('%' || LOWER(:nameFilter) || '%')
                    OR LOWER("A0"."GROUP") LIKE ('%' || LOWER(:nameFilter) || '%'))
```

`select false and false or true` evaluates to `true`. 

In this specific case, the condition would match if only the component's group matches the query, even if it belongs to a different project (`"A0"."PROJECT_ID" = :projectId` is `false`).

Adding parens like this: `select false and (false or true)` would evaluate to `false` as expected. The original PR had these parentheses.

nscuro (changes requested):

nscuro (approved):

PR #855, opened 2024-08-15
Teams Column - ApiServer

jmayer-lm:
### Description
Frontend changes [PR 120](https://github.com/DependencyTrack/hyades-frontend/pull/120)
Made changes to ProjectDao.java, Concise.java and ProjectResource.java to add Teams to the schema model. Also, added new query language to find/sort teams when clicked on.

### Addressed Issue
Fixes [issue 1435](https://github.com/DependencyTrack/hyades/issues/1435)

### Checklist


nscuro:
@jmayer-lm You will need to update some tests in `ProjectResourceTest` since the response will now include additional fields, for example this one: https://github.com/DependencyTrack/hyades-apiserver/blob/b6c907445e15a521d02eee9e0541ccb53b4c02b3/src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java#L315-L352

Separately, I think it would be good to have a test that covers searching by team.

src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
```diff
@@ -108,19 +108,22 @@ public enum ConfigPropertyConstants {
...
     private final String description;
+    private final String
     private final ConfigPropertyAccessMode accessMode;
 
     ConfigPropertyConstants(final String groupName,
                             final String propertyName,
                             final String defaultPropertyValue,
                             final PropertyType propertyType,
                             final String description,
+                            final String teams,
```
nscuro, line 119: This will break all enum values in this file, since they don't provide this constructor parameter.

I'd suggest to remove this change, since assigning projects to teams is done in other ways, see comments in the test file.

src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
```diff
@@ -319,7 +319,8 @@ public void getProjectsConciseWithAclTest() {
                 ACCESS_MANAGEMENT_ACL_ENABLED.getPropertyName(),
                 "true",
                 ACCESS_MANAGEMENT_ACL_ENABLED.getPropertyType(),
-                ACCESS_MANAGEMENT_ACL_ENABLED.getDescription()
+                ACCESS_MANAGEMENT_ACL_ENABLED.getDescription(),
+                ACCESS_MANAGEMENT_ACL_ENABLED.getTeams()
```
nscuro, line 323: See comment in `ConfigPropertyConstants`, this doesn't do what you want.

src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
```diff
@@ -502,6 +503,53 @@ public void getProjectsConciseFilterByTagTest() {
...
+        final var projectA = new Project();
+        projectA.setName("acme-app-a");
+        qm.persist(projectA);
+
+        final var projectB = new Project();
+        projectB.setName("acme-app-b");
+        qm.persist(projectB);
+
+        qm.bind(projectB, List.of(qm.createTag("foo")));
+
```
nscuro, line 517: Three things missing here:

1. You need to enable the portfolio ACL feature for this test
2. You need to assign `projectB` to the team of the test user
3. To get the assertion in the second request of this test to work, you need to filter for `team.getName()` (`team` is available in every test method already, and it's the team you authenticate as by using `.header(X_API_KEY, apiKey)`).

The former you can achieve like this: https://github.com/DependencyTrack/hyades-apiserver/blob/5de4be3c7ba133758859ffdfd5874ce9e90a7ff5/src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java#L317-L323

I recommend putting this at the very top of your this test method.

The latter you do like this: https://github.com/DependencyTrack/hyades-apiserver/blob/5de4be3c7ba133758859ffdfd5874ce9e90a7ff5/src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java#L333-L334

src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
```diff
@@ -803,7 +851,9 @@ public void getProjectChildrenConciseWithAclTest() {
                 ACCESS_MANAGEMENT_ACL_ENABLED.getPropertyName(),
                 "true",
                 ACCESS_MANAGEMENT_ACL_ENABLED.getPropertyType(),
-                ACCESS_MANAGEMENT_ACL_ENABLED.getDescription()
+                ACCESS_MANAGEMENT_ACL_ENABLED.getDescription(),
+                ACCESS_MANAGEMENT_ACL_ENABLED.getTeams()
```
nscuro, line 855: See comment in `ConfigPropertyConstants`, this doesn't do what you want.

src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
```diff
@@ -1046,6 +1096,60 @@ public void getProjectChildrenConciseFilterByTagTest() {
...
+        childProjectA.setParent(parentProject);
+        childProjectA.setName("acme-child-app-a");
+        qm.persist(childProjectA);
+
+        final var childProjectB = new Project();
+        childProjectB.setParent(parentProject);
+        childProjectB.setName("acme-child-app-b");
+        qm.persist(childProjectB);
+
+        qm.bind(childProjectB, List.of(qm.createTeam("foo")));
```
nscuro, line 1115: See comment above, teams are associated with projects like this: https://github.com/DependencyTrack/hyades-apiserver/blob/5de4be3c7ba133758859ffdfd5874ce9e90a7ff5/src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java#L333-L334

`qm.bind` currently only works for tags.

nscuro (changes requested):

src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
jmayer-lm: For clarification,  my team and I believe that this can be achieved similar to lines 141-143 of ProjectResourceTest.java?
```suggestion
// Create project and give access to current principal's team.
        final Project accessProject = qm.createProject("acme-app-a", null, "1.0.0", null, null, null, true, false);
        accessProject.setAccessTeams(List.of(team));
```

nscuro: Yeah, those lines are achieving the same. For the tests, it doesn't really matter if you use `Project.setAccessTeams(listOfTeams)` or `Project.addAccessTeam(team)`. Although the former is a bit less verbose IMO.

src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
```diff
@@ -502,6 +502,57 @@ public void getProjectsConciseFilterByTagTest() {
...
+                .get();
+        assertThat(response.getStatus()).isEqualTo(200);
+        assertThat(response.getHeaderString(TOTAL_COUNT_HEADER)).isEqualTo("1");
+        assertThatJson(getPlainTextBody(response)).isEqualTo("""
+                [
+                  {
+                    "uuid": "${json-unit.any-string}",
+                    "name": "acme-app-b",
+                    "active": true,
+                    "taeam": [
```
nscuro, line 545: Looks like a typo sneaked in here :detective: 

```suggestion
                    "teams": [
```

jmayer-lm: Thanks for that catch!

jmayer-lm:
@nscuro would ".formatted(team.getName()));" allieviate the 3 failed tests? Unsure if its a formatting issue as to why different outcomes are showing OR if it pertains to how the project is being created/accessed. Thoughts?

src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
```diff
@@ -1046,6 +1097,60 @@ public void getProjectChildrenConciseFilterByTagTest() {
...
+        assertThat(response.getHeaderString(TOTAL_COUNT_HEADER)).isEqualTo("1");
+        assertThatJson(getPlainTextBody(response)).isEqualTo("""
+                [
+                  {
+                    "uuid": "${json-unit.any-string}",
+                    "name": "acme-child-app-b",
+                    "active": true,
+                    "teams": [
+                      {
+                        "name": "foo"
```
nscuro, line 1144: The format string `%s` is missing here:

```suggestion
                        "name": "%s"
```

nscuro:
@jmayer-lm Two existing tests are failing now because the API responses now contain the `teams` field. You will need to extend the expected JSON for these tests:

* `getProjectChildrenConciseWithAclTest`
* `getProjectsConciseWithAclTest`

The build logs provide a detailed explanation of what's missing:

```
Different keys found in node "[0]", extra: "[0].teams", expected: <{"active":true,"hasChildren":false,"name":"acme-app-a","uuid":"${json-unit.any-string}"}> but was: <{"active":true,"hasChildren":false,"name":"acme-app-a","teams":[{"name":"Test Users"}],"uuid":"5125a66b-0969-495d-99f8-37ecca09d918"}>
```

So yeah, following the same `.formatted` approach as with the new tests should work.

src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
```diff
@@ -348,7 +348,7 @@ public void getProjectsConciseWithAclTest() {
                     "hasChildren": false
                   }
                 ]
-                """);
+                """.formatted(team.getName()));
```
nscuro, line 351: You still need to extend the JSON to include the `teams` field, just like you did here: https://github.com/DependencyTrack/hyades-apiserver/pull/855/files#diff-24f2f9ee52fee0de1bc5c70446b58dd249f9e0c0c6a24514d5dddf028b010589R1142-R1146

nscuro (changes requested):

src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
jmayer-lm: Thank for you pointing that out

nscuro (approved):
Thanks @jmayer-lm!

PR #856, opened 2024-08-15
Use native UUID type for UUID colums

sahibamittal:
### Description

Replace `UUID` columns type from `VARCHAR` to Postgres native `UUID`.

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/1417

### Checklist

### DIFF
--- a/src/main/resources/migration/changelog-v5.5.0.xml
+++ b/src/main/resources/migration/changelog-v5.5.0.xml
@@ -170,4 +170,28 @@
         <addDefaultValue tableName="VULNERABILITY_POLICY" columnName="OPERATION_MODE" defaultValue="APPLY"/>
         <addNotNullConstraint tableName="VULNERABILITY_POLICY" columnName="OPERATION_MODE"/>
     </changeSet>
+
+    <changeSet id="v5.5.0-15" author="sahibamittal">
+        <modifyDataType tableName="AFFECTEDVERSIONATTRIBUTION" columnName="UUID" newDataType="UUID"/>
+        <modifyDataType tableName="BOM" columnName="UUID" newDataType="UUID"/>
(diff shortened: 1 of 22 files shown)


src/main/resources/migration/changelog-v5.5.0.xml
```diff
@@ -170,4 +170,29 @@
         <addDefaultValue tableName="VULNERABILITY_POLICY" columnName="OPERATION_MODE" defaultValue="APPLY"/>
         <addNotNullConstraint tableName="VULNERABILITY_POLICY" columnName="OPERATION_MODE"/>
     </changeSet>
+
+    <changeSet id="v5.5.0-15" author="sahibamittal">
```
nscuro, line 174: Since this would be for the `5.6.0` release, consider moving this to a new changelog file as per https://dependencytrack.github.io/hyades/latest/development/database-migrations/#adding-migrations

src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
```diff
@@ -801,7 +801,7 @@ public interface Dao {
                 INNER JOIN
                   "PROJECT" AS "P" ON "P"."ID" = "C"."PROJECT_ID"
                 WHERE
-                  "C"."UUID" = (:uuid)::TEXT
+                  "C"."UUID" = (:uuid)::UUID
```
nscuro, line 804: Can you try if omitting the cast (`::UUID`) also works? It *should*, and if it does, I think we should omit it to make the queries less verbose.

For background, the casting was previously necessary because JDBI mapped Java `UUID`s to PostgreSQL `UUID`.

nscuro (approved):

PR #866, opened 2024-08-22
Port: Replace author with authors

2000rosser:
### Description
`author` field is deprecated in CycloneDX 1.6, and `authors` has been introduced.

### Addressed Issue
[#3969](https://github.com/DependencyTrack/dependency-track/pull/3969)

Relates to: https://github.com/DependencyTrack/hyades/issues/1358

### Checklist


src/main/resources/migration/changelog-v5.3.0.xml
nscuro, line 1: Existing changelog entries must not be modified. You'll need to add a changeset to `changelog-v5.6.0.xml` that adds the new column and drops the old one. We have a bit of documentation for this here: https://dependencytrack.github.io/hyades/latest/development/database-migrations/

src/test/java/org/dependencytrack/persistence/converter/ToolsJsonConverterTest.java
```diff
@@ -198,7 +206,11 @@ public void testConvertToAttribute() {
                 {
                   "components": [
                     {
-                       "author": "componentAuthor",
```
nscuro, line 201: For the sake of backward-compatibility, we should leave this `author` field here. Since existing records in the database will have `author` populated, but not `authors`. We should verify that that won't blow up.

nscuro (changes requested):
The original PR (https://github.com/DependencyTrack/dependency-track/pull/3969) had a few more tests, e.g. in `ComponentResourceTest`, `BomResourceTest`. Any chance you can add them here, too?

src/main/resources/migration/changelog-v5.6.0.xml
```diff
@@ -33,4 +33,9 @@
         <modifyDataType tableName="VULNERABLESOFTWARE" columnName="UUID" newDataType="UUID"/>
         <modifyDataType tableName="WORKFLOW_STATE" columnName="TOKEN" newDataType="UUID"/>
     </changeSet>
+
+    <changeSet id="v5.6.0-2" author="rossmurphy974@gmail.com">
+        <renameColumn tableName="PROJECT" oldColumnName="AUTHOR" newColumnName="AUTHORS"/>
+        <renameColumn tableName="COMPONENT" oldColumnName="AUTHOR" newColumnName="AUTHORS"/>
```
nscuro, line 39: Renaming the column will break deserialization, since `AUTHORS` is expected to contain JSON, but `AUTHOR` was not a JSON field.

I think a better approach would be:

1. Create the new `AUTHORS` column
2. Populate `AUTHORS` based on values in the old `AUTHOR` column
3. Drop the `AUTHOR` column

For step (2), you could use the [`sql`](https://docs.liquibase.com/change-types/sql.html) change type like this (not tested):

```sql
UPDATE "PROJECT"
   SET "AUTHORS" = JSON_BUILD_ARRAY(JSON_BUILD_OBJECT('name', "AUTHOR"))::TEXT
 WHERE "AUTHOR" IS NOT NULL
```

See [here](https://www.postgresql.org/docs/12/functions-json.html) for details on the JSON functions. You can find other examples of the `sql` change types in changelog files for prior versions.

nscuro (approved):

PR #870, opened 2024-08-26
Revert "Revert "Use native `UUID` type for UUID colums""

sahibamittal:
Reverts DependencyTrack/hyades-apiserver#867


nscuro (approved):
Reverting the revert... Let's hope we don't have to revert again.

image

PR #886, opened 2024-09-03
Improve Vulnerability query endpoints performance

sahibamittal:
### Description

Replace query fetching in for loop by improved SQL queries for endpoints `/api/v1/vulnerability/component/{uuid}` and `/api/v1/vulnerability/project/{uuid}`. 

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/1424

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
@@ -356,10 +356,10 @@ final class VulnerabilityQueryManager extends QueryManager implements IQueryMana
     @RegisterBeanMapper(Vulnerability.class)
     public PaginatedResult getVulnerabilities(Component component, boolean includeSuppressed) {
         final List<Vulnerability> componentVulnerabilities = withJdbiHandle(
-                handle -> handle.attach(VulnerabilityDao.class).getVulnerabilitiesByComponent(component.getId()));
+                handle -> handle.attach(VulnerabilityDao.class).getVulnerabilitiesByComponent(component.getId(), includeSuppressed));
         final List<VulnerabilityProjectCountRow> vulnerabilityProjectCountList = withJdbiHandle(
                 handle -> handle.attach(VulnerabilityDao.class).getVulnerabilityProjectCount(
-                        componentVulnerabilities.stream().map(Vulnerability::getId).toList(), includeSuppressed));
+                        componentVulnerabilities.stream().map(Vulnerability::getId).toList()));
         for (var vulnerabilityProjectCount : vulnerabilityProjectCountList) {
             componentVulnerabilities.stream().filter(vuln -> vuln.getId() == vulnerabilityProjectCount.id()).findFirst()
                     .ifPresent(vulnerability -> {
--- a/src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
+++ b/src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
@@ -55,9 +55,6 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import static org.dependencytrack.persistence.jdbi.JdbiFactory.openJdbiHandle;
-import static org.dependencytrack.persistence.jdbi.JdbiFactory.withJdbiHandle;
-import static org.dependencytrack.persistence.jdbi.mapping.RowMapperUtil.longArray;
-import static org.dependencytrack.persistence.jdbi.mapping.RowMapperUtil.maybeSet;
 import static org.dependencytrack.util.PrincipalUtil.getPrincipalTeamIds;
 
 final class VulnerabilityQueryManager extends QueryManager implements IQueryManager {
@@ -462,7 +459,7 @@ final class VulnerabilityQueryManager extends QueryManager implements IQueryMana
             final var dao = jdbiHandle.attach(VulnerabilityDao.class);
             projectVulnerabilities = dao.getVulnerabilitiesByProject(project.getId(), includeSuppressed);
             dao.getVulnerableComponents(project.getId(), projectVulnerabilities.stream().map(Vulnerability::getId).toList())
-                    .stream().map(vc -> vulnerableComponents.put(vc.getId(), vc));
+                    .stream().forEach(vc -> vulnerableComponents.put(vc.getId(), vc));
         }
         for (var projectVulnerability : projectVulnerabilities) {
             if (projectVulnerability.getComponents() != null) {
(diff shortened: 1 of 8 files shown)


src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
```diff
@@ -461,23 +451,10 @@ public long getVulnerabilityCount(Project project, boolean includeSuppressed) {
...
-        final List<Vulnerability> vulnerabilities = new ArrayList<>();
-        final List<Component> components = getAllComponents(project);
-        for (final Component component: components) {
-            final Collection<Vulnerability> componentVulns = pm.detachCopyAll(
-                    getAllVulnerabilities(component, includeSuppressed)
-            );
-            Map<String, Epss> matchedEpssList = getEpssForCveIds(
-                    componentVulns.stream().map(vuln -> vuln.getVulnId()).distinct().toList());
-            for (final Vulnerability componentVuln: componentVulns) {
-                componentVuln.setComponents(Collections.singletonList(pm.detachCopy(component)));
```
nscuro, line 474: Note how in the original logic, vulnerable components are attached to the vulnerability. And the same vulnerability could appear multiple times in the result, if it affects multiple components.

In the new SQL query, `DISTINCT` is used to de-duplicate based on vulnerability attributes, and no component information is returned at all.

nscuro (changes requested):
There are some differences in data returned by the REST API before and after this change.

I'd recommend to update the corresponding tests in [`VulnerabilityResourceTest`](https://github.com/DependencyTrack/hyades-apiserver/blob/main/src/test/java/org/dependencytrack/resources/v1/VulnerabilityResourceTest.java) to using `assertThatJson`. That will allow you to ensure that the response content before and after this change is identical.

src/test/java/org/dependencytrack/resources/v1/VulnerabilityResourceTest.java
```diff
@@ -853,9 +853,9 @@ private class SampleData {
             c2 = new Component();
             c2.setProject(p1);
             c2.setName("Component 2");
-            c1.setMd5("5EABD62FA03D159A96C77E6EEB6C7027");
-            c1.setSha1("108BCF94A1C0E0F915B935C97F6BB9E50FB7C246");
-            c1.setSha256("418716B003FE0268B6521EF7ACBED13F5BA491D593896D5DEB2058C42D87002D");
+            c2.setMd5("5EABD62FA03D159A96C77E6EEB6C7027");
+            c2.setSha1("108BCF94A1C0E0F915B935C97F6BB9E50FB7C246");
+            c2.setSha256("418716B003FE0268B6521EF7ACBED13F5BA491D593896D5DEB2058C42D87002D");
```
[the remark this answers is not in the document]
sahibamittal: yes I realised hashes were being set for c1 twice and these are supposed to be set for c2.

src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
```diff
@@ -462,22 +453,22 @@ public long getVulnerabilityCount(Project project, boolean includeSuppressed) {
...
+            final var dao = jdbiHandle.attach(VulnerabilityDao.class);
+            projectVulnerabilities = dao.getVulnerabilitiesByProject(project.getId(), includeSuppressed);
+            dao.getVulnerableComponents(project.getId(), projectVulnerabilities.stream().map(Vulnerability::getId).toList())
+                    .stream().forEach(vc -> vulnerableComponents.put(vc.getId(), vc));
+        }
+        vulnerableComponents.values().forEach(vc -> vc.setProject(project));
+        for (var projectVulnerability : projectVulnerabilities) {
+            if (projectVulnerability.getComponents() != null) {
+                projectVulnerability.setComponents(
+                        projectVulnerability.getComponents().stream().map(c -> vulnerableComponents.get(c.getId())).toList());
```
nscuro, line 468: This will assign components vulnerable to **any** of the vulnerabilities, not only this particular `projectVulnerability`.

Since `dao.getVulnerableComponents` fetches components for all vulnerabilities in bulk, you're loosing context of which vulnerability a returned component is for.

You'll need a marker of sorts to tell you which component was returned for which vulnerability. One option is to additionally select `"VULNERABILITY"."ID"` in the query, and then collecting the results into a `Map<Long, List<Component>> componentsByVulnId` structure.

The above would then allow you to call this in the `for` loop:

```java
projectVulnerability.setComponents(componentsByVulnId.get(projectVulnerability.getId());
```

sahibamittal: List of components are fetched for each vulnerability in SQL query of `getVulnerabilitiesByProject` -> `ARRAY_AGG(DISTINCT("COMPONENT"."ID")) as "componentIdsArray"`.
Then based on this array associated with each vulnerability, components are set for those component ids `projectVulnerability.getComponents().stream().map(c -> vulnerableComponents.get(c.getId())).toList());`

sahibamittal: Thanks Niklas!

src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java
```diff
@@ -462,22 +453,22 @@ public long getVulnerabilityCount(Project project, boolean includeSuppressed) {
...
-                componentVuln.setEpss(matchedEpssList.get(componentVuln.getVulnId()));
+        List<Vulnerability> projectVulnerabilities;
+        var vulnerableComponents = new HashMap<Long, Component>();
+        try (final Handle jdbiHandle = openJdbiHandle((this.request))) {
+            final var dao = jdbiHandle.attach(VulnerabilityDao.class);
+            projectVulnerabilities = dao.getVulnerabilitiesByProject(project.getId(), includeSuppressed);
+            dao.getVulnerableComponents(project.getId(), projectVulnerabilities.stream().map(Vulnerability::getId).toList())
+                    .stream().forEach(vc -> vulnerableComponents.put(vc.getId(), vc));
+        }
+        vulnerableComponents.values().forEach(vc -> vc.setProject(project));
```
nscuro, line 464: Seeing as the project information is repeated in the API response for every component, I wonder if we should just omit this. I think the inclusion of it was more of a bug in the previous implementation than an intentional inclusion.

Since clients are requesting vulnerabilities for a specific project already, do we really need to include that project's details in the response again?

I do realize this would technically be a breaking change, but I can't fathom anyone actually needing this kind of nested and redundant data. Thoughts?

sahibamittal: yeah I completely agree, there is no point of nested project info in the component data when it is requested of specific project. I'll remove this part 👍🏼

nscuro (approved):

PR #914, opened 2024-09-18
Port : Support inclusion/exclusion of projects from BOM validation with tags

sahibamittal:
### Description

Supports the inclusion or exclusion of projects from BOM validation, using tags.

- Removes the artifact:bom.validation.enabled config property

- Adds the new artifact:bom.validation.mode config property

  - Valid options are ENABLED, DISABLED, ENABLED_FOR_TAGS, DISABLED_FOR_TAGS
  - Existing artifact:bom.validation.enabled configs are migrated to their artifact:bom.validation.mode counterpart

- Adds the new artifact:bom.validation.tags.exclusive and artifact:bom.validation.tags.inclusive config properties to support the new ENABLED_FOR_TAGS and DISABLED_FOR_TAGS modes

### Addressed Issue

Port change https://github.com/DependencyTrack/hyades/issues/1358

### Checklist


nscuro (changes requested):
The original PR had a database migration that is missing here. The migration [sets the `bom.validation.mode` property based on the existing `bom.validation.enabled` value](https://github.com/DependencyTrack/dependency-track/pull/4109/files#diff-ea625c4482c2c1c109c2a4ea206160c9761c0ab3701cbe4afcf33aea795c28d3R79-R119), and [deletes the latter](https://github.com/DependencyTrack/dependency-track/pull/4109/files#diff-ea625c4482c2c1c109c2a4ea206160c9761c0ab3701cbe4afcf33aea795c28d3R121-R131) afterwards.

I think we can skip the first part since Hyades is not yet GA, but at the very least we should delete the old property.

nscuro (approved):

PR #970, opened 2024-11-11
Quick fix : component author column mapping 

sahibamittal:
### Description

Fixes the column mapping exception for Component -> Authors

### Addressed Issue

NA

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/jdbi/mapping/OrganizationalContactMapper.java
+++ b/src/main/java/org/dependencytrack/persistence/jdbi/mapping/OrganizationalContactMapper.java
@@ -25,7 +25,6 @@ import org.jdbi.v3.core.statement.StatementContext;
 
 import java.sql.ResultSet;
 import java.sql.SQLException;
-import java.util.ArrayList;
 import java.util.List;
 
 public class OrganizationalContactMapper implements ColumnMapper<List<OrganizationalContact>> {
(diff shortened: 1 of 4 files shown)


src/main/java/org/dependencytrack/persistence/jdbi/mapping/OrganizationalContactMapper.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class OrganizationalContactMapper implements ColumnMapper<List<OrganizationalContact>> {
+
+    @Override
+    public List<OrganizationalContact> map(ResultSet r, int columnNumber, StatementContext ctx) throws SQLException {
+        if (r.getBytes(columnNumber) == null) {
+            return new ArrayList<>();
+        }
```
nscuro, line 37: I think you can pass the column value directly to the converter. It already handles `null` values: 
https://github.com/DependencyTrack/hyades-apiserver/blob/68f786b699946c2badf231b567aba2c53ea7b84d/src/main/java/org/dependencytrack/persistence/converter/AbstractJsonConverter.java#L75-L77

Also note how the converter returns `null` for `null` inputs, while this mapper would return an empty list, yielding a different behavior. Better to follow what the converter is doing.

nscuro (changes requested):

nscuro (approved):

PR #971, opened 2024-11-18
Quick-fix : sql query to add tags while fetching vulnerabilities

sahibamittal:
### Description

While fetching vulnerabilities, the sql query did not include the tags, causing empty list in frontend (vulnerability modal) and endpoints.

### Addressed Issue

NA

### Checklist


src/main/java/org/dependencytrack/persistence/jdbi/VulnerabilityDao.java
```diff
@@ -78,6 +78,12 @@ public interface VulnerabilityDao {
                  , JSONB_VULN_ALIASES("V"."SOURCE", "V"."VULNID") AS "vulnAliasesJson"
                  , "EPSS"."SCORE"
                  , "EPSS"."PERCENTILE"
+                 ,  (SELECT JSONB_AGG(JSONB_BUILD_OBJECT('id', "ID", 'name', "NAME"))
```
nscuro, line 81: Selecting `ID` seems to be unnecessary as it's excluded in the JSON output. Would `ARRAY_AGG("TAG"."NAME")` be a better option? Like so:
https://github.com/DependencyTrack/hyades-apiserver/blob/682427bbb4c7d7d520c82c08bd0dccad682b0b0b/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java#L62-L66

sahibamittal: The `tags` is a list of string in record `ConciseProjectListRow` in `ProjectDao`.
For `Vulnerability` object mapper, we need list of `Tag` objects for which `JSONB_BUILD_OBJECT` is making it easier.

PR #984, opened 2024-11-27
Replace update workflow state methods with DAO queries

sahibamittal:
### Description

Replace update workflow state method with DAO queries + remove the method not in use.

### Addressed Issue

NA

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/tasks/BomUploadProcessingTask.java
+++ b/src/main/java/org/dependencytrack/tasks/BomUploadProcessingTask.java
@@ -65,7 +65,6 @@ import org.dependencytrack.persistence.QueryManager;
 import org.dependencytrack.persistence.jdbi.WorkflowDao;
 import org.dependencytrack.util.InternalComponentIdentifier;
 import org.dependencytrack.util.WaitingLockConfiguration;
-import org.jdbi.v3.core.Handle;
 import org.json.JSONArray;
 import org.slf4j.MDC;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/dependencytrack/tasks/BomUploadProcessingTask.java
```diff
@@ -174,13 +177,15 @@ public void inform(final Event e) {
         try (var ignoredMdcProjectUuid = MDC.putCloseable(MDC_PROJECT_UUID, ctx.project.getUuid().toString());
              var ignoredMdcProjectName = MDC.putCloseable(MDC_PROJECT_NAME, ctx.project.getName());
              var ignoredMdcProjectVersion = MDC.putCloseable(MDC_PROJECT_VERSION, ctx.project.getVersion());
-             var ignoredMdcBomUploadToken = MDC.putCloseable(MDC_BOM_UPLOAD_TOKEN, ctx.token.toString())) {
-            processEvent(ctx, event);
+             var ignoredMdcBomUploadToken = MDC.putCloseable(MDC_BOM_UPLOAD_TOKEN, ctx.token.toString());
+             final Handle jdbiHandle = openJdbiHandle()) {
```
nscuro, line 181: Unfortunately this has the potential to cause some issues, because we already keep a database connection open for the entire duration of `processBom` (since it's running in a single transaction). By additionally [opening a JDBI handle](https://jdbi.org/#_managing_the_handle_lifecycle_manually), we'll be keeping an additional connection open.

I'd recommend to use `useJdbiTransaction` or `inJdbiTransaction` ad-hoc, whenever you need to update the workflow state. That ensures that the connection used for those operations is released quickly.

src/main/java/org/dependencytrack/tasks/BomUploadProcessingTask.java
```diff
@@ -204,7 +209,8 @@ private void processEvent(final Context ctx, final BomUploadEvent event) {
             return;
         }
 
-        startBomProcessingWorkflowStep(ctx);
+        workflowDao.updateState(WorkflowStep.BOM_CONSUMPTION, ctx.token, WorkflowStatus.COMPLETED, null);
+        workflowDao.startState(WorkflowStep.BOM_PROCESSING, ctx.token);
```
nscuro, line 213: Use `useJdbiTransaction` to ensure that both operations happen atomically.

nscuro (changes requested):
Thanks, the more JDBI the better! :D

Just a note on connection usage, otherwise looks good!

nscuro (approved):

PR #1002, opened 2024-12-13
Migrate project `active` flag to date type `inactiveSince`

sahibamittal:
### Description

In order to implement feature of inactive project deletion, first the `active` flag of `Project` needs to be migrated a Date type field `inactiveSince` to manage the duration of its deactivation.  

### Addressed Issue

Subtask of https://github.com/DependencyTrack/hyades/issues/1600

### Checklist


src/test/resources/migration/custom/schema-v5.2.0-postgresql.sql
nscuro, line 1: This file is meant specifically for testing the migration from v5.2.0, so should remain untouched by this change.

src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
```diff
@@ -741,7 +741,7 @@ public Response patchProject(
                 modified |= setIfDifferent(jsonProject, project, Project::getCpe, Project::setCpe);
                 modified |= setIfDifferent(jsonProject, project, Project::getPurl, Project::setPurl);
                 modified |= setIfDifferent(jsonProject, project, Project::getSwidTagId, Project::setSwidTagId);
-                modified |= setIfDifferent(jsonProject, project, Project::isActive, Project::setActive);
+                modified |= setIfDifferent(jsonProject, project, Project::getInactiveSince, Project::setInactiveSince);
```
nscuro, line 744: This allows clients to choose a custom date, which I think is not what we want. The client should be telling us that a project shall be deactivated or reactivated, but not *when* it was deactivated. The latter is an implementation detail for us, that should not "bleed" through the API.

src/main/java/org/dependencytrack/tasks/RepositoryMetaAnalysisTask.java
```diff
@@ -144,9 +144,9 @@ private void dispatchComponents(final List<ComponentProjection> components) {
 
     private List<ComponentProjection> fetchNextComponentsPage(final PersistenceManager pm, final Project project, final long offset) throws Exception {
         try (final Query<Component> query = pm.newQuery(Component.class)) {
-            var filter = "project.active == :projectActive && purlCoordinates != null";
+            var filter = "project.inactiveSince == :projectInactiveSince && purlCoordinates != null";
             var params = new HashMap<String, Object>();
-            params.put("projectActive", true);
+            params.put("projectInactiveSince", null);
```
nscuro, line 149: `null` should be passed as a literal, see https://github.com/DependencyTrack/dependency-track/issues/2540

```suggestion
            var filter = "project.inactiveSince == null && purlCoordinates != null";
            var params = new HashMap<String, Object>();
```

src/main/java/org/dependencytrack/tasks/VulnerabilityAnalysisTask.java
```diff
@@ -149,9 +149,9 @@ private void dispatchComponents(final UUID scanToken, final List<ComponentProjec
 
     private List<ComponentProjection> fetchNextComponentsPage(final PersistenceManager pm, final Project project, final Long lastId) throws Exception {
         try (final Query<Component> query = pm.newQuery(Component.class)) {
-            var filter = "project.active == :projectActive";
+            var filter = "project.inactiveSince == :projectInactiveSince";
             var params = new HashMap<String, Object>();
-            params.put("projectActive", true);
+            params.put("projectInactiveSince", null);
```
nscuro, line 154: `null` should be passed as a literal, see https://github.com/DependencyTrack/dependency-track/issues/2540

```suggestion
            var filter = "project.inactiveSince == null";
            var params = new HashMap<String, Object>();
```

src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
```diff
@@ -2517,19 +2473,16 @@ public void issue3883RegressionTest() {
...
                   "properties": [],
                   "tags": [],
-                  "active": true,
                   "isLatest": false,
                   "versions": [
                     {
                       "uuid": "${json-unit.any-string}",
-                      "version": "1.0.0",
-                      "active": true
+                      "version": "1.0.0"
```
nscuro, line 2485: The `active` flag is required by the frontend to visually separate active from inactive project versions in the version dropdown: https://github.com/DependencyTrack/hyades-frontend/blob/5094b723e9b09e6f6d1bf4b2ca374b4e25299e61/src/views/portfolio/projects/Project.vue#L42-L67

By not returning this flag, we'd be breaking said frontend feature.

nscuro (commented):
Great work on this @sahibamittal!

I left a comment about clients being able to set `inactiveSince` via API. I believe it would make sense to hide this implementation detail from clients, and continue to expose `active` as a boolean instead. You already had it working like this in a previous commit I think, where you had `isActive` return `inactiveSince == null`.

Some additional mapping would be needed for *setting* this value in the `updateProject` and `patchProject` methods.

We can return `inactiveSince` as an additional, read-only property since it might still be of interest. But it shouldn't be possible to control its value - that's up to us to manage.

You can mark a property as read-only for the OpenAPI spec like this:

https://github.com/DependencyTrack/hyades-apiserver/blob/a844eb8725a2e817a90bbd64abd1de4a469bd297/src/main/java/org/dependencytrack/model/Project.java#L310

WDYT?

src/main/java/org/dependencytrack/model/Project.java
```diff
@@ -325,6 +326,8 @@ public enum FetchGroup {
 
     private transient String author;
 
+    private transient boolean active;
```
nscuro, line 329: I think this field is no longer needed, since `isActive` and `setActive` operate on the `inactiveSince` field.

src/main/java/org/dependencytrack/model/ProjectVersion.java
```diff
@@ -21,11 +21,12 @@
...
 import java.io.Serializable;
+import java.util.Date;
 import java.util.UUID;
 
 /**
  * Value object holding UUID and version for a project
  */
 @JsonInclude(JsonInclude.Include.NON_NULL)
-public record ProjectVersion(UUID uuid, String version, Boolean active) implements Serializable {
+public record ProjectVersion(UUID uuid, String version, Date inactiveSince) implements Serializable {
```
nscuro, line 31: Similar to `Project`, the `boolean active` field should stay for API-facing classes. It can *additionally* include `inactiveSince`.

src/main/java/org/dependencytrack/policy/cel/mapping/ProjectProjection.java
```diff
@@ -41,8 +41,8 @@ public class ProjectProjection {
     @MappedField(sqlColumnName = "CLASSIFIER")
     public String classifier;
 
-    @MappedField(protoFieldName = "is_active", sqlColumnName = "ACTIVE")
-    public Boolean isActive;
+    @MappedField(protoFieldName = "inactive_since", sqlColumnName = "INACTIVE_SINCE")
+    public Date inactiveSince;
```
nscuro, line 45: Hmmm, just checked - it seems this entire class is no longer used. It's referenced in `CelPolicyQueryManager#fetchProject`, but that method is unused. Instead of adjusting this line, you could delete said code instead.

src/main/java/org/dependencytrack/policy/cel/persistence/CelPolicyProjectRowMapper.java
```diff
@@ -51,7 +51,7 @@ public Project map(final ResultSet rs, final StatementContext ctx) throws SQLExc
         maybeSet(rs, "name", ResultSet::getString, builder::setName);
         maybeSet(rs, "version", ResultSet::getString, builder::setVersion);
         maybeSet(rs, "classifier", ResultSet::getString, builder::setClassifier);
-        maybeSet(rs, "is_active", ResultSet::getBoolean, builder::setIsActive);
+        maybeSet(rs, "inactive_since", RowMapperUtil::nullableTimestamp, builder::setInactiveSince);
```
nscuro, line 54: We can avoid a breaking change in the policy Protobuf schema by keeping `is_active`, and populating it based on whether `inactiveSince` is `null`. That would also solve the failing `buf` build check.

src/main/proto/org/dependencytrack/policy/v1/policy.proto
```diff
@@ -90,7 +90,7 @@ message Project {
   string name = 3;
   optional string version = 4;
   optional string classifier = 5;
-  bool is_active = 6;
+  optional google.protobuf.Timestamp inactive_since = 6;
```
nscuro, line 93: See comment in `CelPolicyProjectRowMapper.java`. We should avoid this breaking change.

src/main/java/org/dependencytrack/resources/v1/vo/AffectedProject.java
```diff
@@ -42,12 +43,12 @@ public class AffectedProject {
 
     private final List<UUID> affectedComponentUuids;
 
-    public AffectedProject(UUID uuid, boolean dependencyGraphAvailable, String name, String version, boolean active, List<UUID> affectedComponentUuids) {
+    public AffectedProject(UUID uuid, boolean dependencyGraphAvailable, String name, String version, Instant inactiveSince, List<UUID> affectedComponentUuids) {
```
nscuro, line 46: Similar to `Project`, the `boolean active` field should stay.

src/main/java/org/dependencytrack/resources/v1/vo/ConciseProject.java
```diff
@@ -42,7 +42,7 @@ public record ConciseProject(
         @Schema(description = "Name of the project", requiredMode = Schema.RequiredMode.REQUIRED) String name,
         @Schema(description = "Version of the project") String version,
         @Schema(description = "Classifier of the project") Classifier classifier,
-        @Schema(description = "Whether the project is active", requiredMode = Schema.RequiredMode.REQUIRED) boolean active,
+        @Schema(description = "Timestamp of the project deactivation", type = "number", example = "1719499619599") Date inactiveSince,
```
nscuro, line 45: Similar to `Project`, the `boolean active` field should stay for API-facing classes.

nscuro (commented):
Great work again, thanks! Just a few final suggestions so we can avoid breaking changes.

src/main/java/org/dependencytrack/policy/cel/mapping/ProjectProjection.java
sahibamittal: It is also being used by `CelPolicyDao#loadRequiredFields` to filter sql columns.

nscuro: You're right! My bad!

sahibamittal: I'll delete this field from projection though.

src/test/java/org/dependencytrack/policy/cel/CelPolicyEngineTest.java
```diff
@@ -295,7 +295,6 @@ public void testEvaluateProjectWithAllFields() {
                   && project.name == "projectName"
                   && project.version == "projectVersion"
                   && project.classifier == "APPLICATION"
-                  && project.is_active
```
nscuro, line 298: Can we keep the `is_active` check here please? Since the test's purpose is to verify that *all* fields are loaded correctly. I think it would need to be changed to `!project.is_active` now, since `inactiveSince` is being populated above.

src/test/java/org/dependencytrack/policy/cel/persistence/CelPolicyDaoTest.java
```diff
@@ -109,7 +109,6 @@ public void testLoadRequiredFieldsForProject() throws Exception {
                           "name": "projectName",
                           "version": "projectVersion",
                           "classifier": "APPLICATION",
-                          "isActive": true,
```
nscuro, line 112: If I'm not mistaken this should be `"isActive": false` now, since `inactiveSince` is populated above.

nscuro (approved):
Good stuff, feel free to merge all the PRs when you're ready :shipit:

PR #1024, opened 2025-01-15
Suppress rejected vulnerabilities (records with empty data) by SNYK

sahibamittal:
### Description

When SNYK sends records with empty data, apiserver receives vulnerability with rejected set by vulnerability analyser.
This change is to find SNYK vulnerabilities for the project component and suppress them.

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/1620

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
+++ b/src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
@@ -246,7 +246,8 @@ public class VulnerabilityScanResultProcessor implements Processor<ScanKey, Scan
         final var syncedVulns = new HashSet<Vulnerability>();
 
         for (final org.cyclonedx.proto.v1_6.Vulnerability reportedVuln : scannerResult.getBom().getVulnerabilitiesList()) {
-            if (reportedVuln.getRejected() != null && scannerResult.getScanner() == SCANNER_SNYK) {
+            if (reportedVuln.getRejected() != null && !reportedVuln.getRejected().toString().isEmpty()
+                    && scannerResult.getScanner() == SCANNER_SNYK) {
                 // If vulnerability has rejected field, component is no longer vulnerable by SNYK, suppress it.
                 suppressComponentVulnbySnyk(qm, component);
             } else {
--- a/src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
+++ b/src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
@@ -950,7 +950,6 @@ public class VulnerabilityScanResultProcessor implements Processor<ScanKey, Scan
                       AND "A"."VULNERABILITY_ID" = "CV"."VULNERABILITY_ID"
                     WHERE "CV"."COMPONENT_ID" = :componentId
                       AND "V"."SOURCE" = 'SNYK'
-                      AND NOT "A"."SUPPRESSED"
                      <#if vulnIdsToExclude>
                        AND "V"."VULNID" != ANY(:vulnIdsToExclude)
                      </#if>
@@ -966,6 +965,7 @@ public class VulnerabilityScanResultProcessor implements Processor<ScanKey, Scan
                  -- Update existing record if not already suppressed.
                  ON CONFLICT ("PROJECT_ID", "COMPONENT_ID", "VULNERABILITY_ID")
                  DO UPDATE SET "SUPPRESSED" = TRUE, "STATE" = 'FALSE_POSITIVE'
+                 WHERE NOT "ANALYSIS"."SUPPRESSED"
                  RETURNING "ID"
             """)
         @DefineNamedBindings
--- a/src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
+++ b/src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
@@ -206,7 +206,7 @@ public class VulnerabilityScanResultProcessor implements Processor<ScanKey, Scan
             return;
         }
 
-        final Set<Vulnerability> syncedVulns = syncVulnerabilities(qm, scanKey, scannerResult);
+        final Set<Vulnerability> syncedVulns = syncVulnerabilities(qm, component, scanKey, scannerResult);
         LOGGER.debug("Synchronized %d vulnerabilities reported by %s for %s (scanKey: %s)"
                 .formatted(syncedVulns.size(), scannerResult.getScanner(), scanKey.getComponentUuid(), prettyPrint(scanKey)));
 
@@ -243,11 +243,8 @@ public class VulnerabilityScanResultProcessor implements Processor<ScanKey, Scan
      * @param scannerResult The {@link ScannerResult} to synchronize vulnerabilities from
      * @return A {@link Set} of synchronized {@link Vulnerability}s
      */
-    private Set<Vulnerability> syncVulnerabilities(final QueryManager qm, final ScanKey scanKey, final ScannerResult scannerResult) {
+    private Set<Vulnerability> syncVulnerabilities(final QueryManager qm, final Component component, final ScanKey scanKey, final ScannerResult scannerResult) {
         final var syncedVulns = new HashSet<Vulnerability>();
-
-        final Component component = withJdbiHandle(handle ->
-                handle.attach(Dao.class).getComponentByUuid(UUID.fromString(scanKey.getComponentUuid())));
 
         if (scannerResult.getScanner() == SCANNER_SNYK && component.hasSnykVulns()) {
             // Compare component's Snyk vulnerabilities and suppress those which are longer vulnerable by SNYK.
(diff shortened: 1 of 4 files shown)


src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
```diff
@@ -904,6 +926,18 @@ ON CONFLICT ("VULNERABILITY_ID", "COMPONENT_ID", "PROJECT_ID") DO UPDATE
...
+                    SET "SUPPRESSED" = true, "STATE" = 'NOT_AFFECTED'
+                    WHERE "COMPONENT_ID" = :componentId
+                    AND "VULNERABILITY_ID" IN
+                        ( SELECT "VULNERABILITY"."ID"
+                            FROM "VULNERABILITY"
+                            WHERE "VULNERABILITY"."SOURCE" = 'SNYK')
+                    RETURNING "ANALYSIS"."ID"
+            """)
+        @RegisterBeanMapper(org.dependencytrack.model.Analysis.class)
+        List<Long> suppressSnykAnalyses(@Bind final long componentId);
```
nscuro, line 940: This suppresses all Snyk vulnerabilities of a component, but what if only one out of `N` vulnerabilities were rejected? Lots of components have multiple vulnerabilities.

nscuro: Also consider that no analysis may exist already, so an insert would be necessary rather than an update.

sahibamittal: If Snyk is reporting no results for a component purl then it should be that Snyk doesn't have any vulnerabilities for the component.
Yes, we can insert new analysis instead of update 👍🏼

nscuro: > If Snyk is reporting no results for a component purl then it should be that Snyk doesn't have any vulnerabilities for the component.

Yeah that's fair. But what if Snyk still reports vulnerabilities, but only a *subset* of what it had previously reported?

sahibamittal: At apiserver end, we can compare the incoming list of vulns for component with it's existing Snyk vulns and suppress the ones which are no longer being reported?

src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
```diff
@@ -904,6 +927,24 @@ ON CONFLICT ("VULNERABILITY_ID", "COMPONENT_ID", "PROJECT_ID") DO UPDATE
...
+                    WHERE "COMPONENT_ID" = :componentId
+                    AND "SUPPRESSED" = 'false'
+                    AND "VULNERABILITY_ID" IN
+                        ( SELECT "VULNERABILITY"."ID"
+                            FROM "VULNERABILITY"
+                            WHERE "VULNERABILITY"."SOURCE" = 'SNYK'
+                            <#if vulnIdsToExclude>
+                                AND "VULNERABILITY"."VULNID" != ANY(:vulnIdsToExclude)
+                            </#if>
+                        )
```
nscuro, line 942: Consider expressing this as a join instead. `UPDATE` has the `FROM` clause for that purpose. For example:

```sql
UPDATE "ANALYSIS"
   SET "SUPPRESSED" = TRUE
     , "STATE" = 'NOT_AFFECTED'
  FROM "VULNERABILITY"
 WHERE "ANALYSIS"."COMPONENT_ID" = :componentId
   AND "ANALYSIS"."VULNERABILITY_ID" = "VULNERABILITY"."ID"
   AND "VULNERABILITY"."SOURCE" = 'SNYK'
<#if vulnIdsToExclude>
   AND "VULNERABILITY"."VULNID" != ANY(:vulnIdsToExclude)
</#if>
```

If `vulnIdsToExclude` would be `null`, this subquery:

```sql
SELECT "VULNERABILITY"."ID"
  FROM "VULNERABILITY"
 WHERE "VULNERABILITY"."SOURCE" = 'SNYK'
```

would yield a rather large result set, which you can avoid with the join as shown above.

src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
```diff
@@ -237,13 +243,18 @@ private void processScannerResult(final QueryManager qm, final Component compone
...
      * @return A {@link Set} of synchronized {@link Vulnerability}s
      */
-    private Set<Vulnerability> syncVulnerabilities(final QueryManager qm, final ScanKey scanKey, final ScannerResult scannerResult) {
+    private Set<Vulnerability> syncVulnerabilities(final QueryManager qm, final Component component, final ScanKey scanKey, final ScannerResult scannerResult) {
         final var syncedVulns = new HashSet<Vulnerability>();
 
+        if (scannerResult.getScanner() == SCANNER_SNYK) {
+            // Compare component's Snyk vulnerabilities and suppress those which are longer vulnerable by SNYK.
+            compareAndSuppressVulnBySnyk(qm, component, scannerResult.getBom().getVulnerabilitiesList());
+        }
```
nscuro, line 252: It would be good to avoid this work if the component does not currently have any vulnerabilities. The *vast* majority of components will not be vulnerable, so there won't ever be anything to suppress.

One way you could achieve this is by extending the `getComponentByUuid` query: https://github.com/DependencyTrack/hyades-apiserver/blob/e312cb4415e659d9287ed7b02006acfdf0ff69cb/src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java#L792-L806

In the simplest scenario, you'd want to know whether the component has any Snyk vulnerabilities:

```sql
SELECT
  "C"."ID"   AS "id",
  "C"."UUID" AS "uuid",
  "P"."ID"   AS "projectId",
  "P"."UUID" AS "projectUuid"
  (SELECT EXISTS(SELECT 1
                   FROM "COMPONENTS_VULNERABILITIES" AS "CV"
                  INNER JOIN "VULNERABILITY" AS "V"
                     ON "V"."ID" = "CV"."VULNERABILITY_ID"
                  WHERE "CV"."COMPONENT_ID" = "C"."ID"
                    AND "V"."SOURCE" = 'SNYK')) AS "hasSnykVulns"
FROM
  "COMPONENT" AS "C"
INNER JOIN
  "PROJECT" AS "P" ON "P"."ID" = "C"."PROJECT_ID"
WHERE
  "C"."UUID" = :uuid
```

You could then check for it:

```java
if (scannerResult.getScanner() == SCANNER_SNYK && component.hasSnykVulns()) {
```

A more future-proof solution might be to fetch the source-vulnId pairs of *all* vulnerabilities assigned to the component. Ideally, the auto-suppression logic should work for other scanners as well, not just Snyk. But for the moment, the `hasSnykVulns` way should suffice.

src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
```diff
@@ -904,6 +927,24 @@ ON CONFLICT ("VULNERABILITY_ID", "COMPONENT_ID", "PROJECT_ID") DO UPDATE
                 """)
         void createAnalysisComments(@BindMethods final Iterable<AnalysisComment> comment);
 
+        @SqlQuery("""
+                UPDATE "ANALYSIS"
```
nscuro, line 931: This still only updates an existing analysis, but doesn't create a new one if necessary.

src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
```diff
@@ -904,6 +927,24 @@ ON CONFLICT ("VULNERABILITY_ID", "COMPONENT_ID", "PROJECT_ID") DO UPDATE
...
+                    SET "SUPPRESSED" = true, "STATE" = 'NOT_AFFECTED'
```
nscuro, line 932: Wondering if perhaps `FALSE_POSITIVE` would be a better state to assign. `NOT_AFFECTED` still suggests that the vulnerability itself is valid, which in this case it wouldn't be.

src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
sahibamittal: If I annotate it with `@SqlUpdate` it should return void, boolean, or Number but I need list of analysis to be returned.

src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
sahibamittal: Good catch, thanks!

src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java
```diff
@@ -240,10 +246,18 @@ private void processScannerResult(final QueryManager qm, final Component compone
     private Set<Vulnerability> syncVulnerabilities(final QueryManager qm, final ScanKey scanKey, final ScannerResult scannerResult) {
         final var syncedVulns = new HashSet<Vulnerability>();
 
+        final Component component = withJdbiHandle(handle ->
```
nscuro, line 249: Any reason for re-fetching the component that was already fetched earlier? https://github.com/DependencyTrack/hyades-apiserver/blob/a8a454a6270bb9cb6163397b7d948d3b2f4499cf/src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java#L170

nscuro (approved):

PR #1042, opened 2025-01-28
Port : Fix REST endpoints for adding tags

sahibamittal:
### Description

Fixes REST endpoints for adding tags.

### Addressed Issue

Ports https://github.com/DependencyTrack/dependency-track/pull/4543
Port change https://github.com/DependencyTrack/hyades/issues/1358

### Checklist


nscuro (dismissed):

nscuro:
@sahibamittal The test is failing because this condition:

https://github.com/DependencyTrack/hyades-apiserver/blob/6f4ed40076eac3cd42c6ad172adfae30e13a969e/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java#L647-L652

can no longer happen in this location. 

The unique constraint violation originating from the database will only be thrown when the DB transaction commits.

Because the statement in question is also wrapped in a `callInTransaction`:

https://github.com/DependencyTrack/hyades-apiserver/blob/6f4ed40076eac3cd42c6ad172adfae30e13a969e/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java#L607

the transaction will only commit after that call completes, so technically [here](https://github.com/DependencyTrack/hyades-apiserver/blob/6f4ed40076eac3cd42c6ad172adfae30e13a969e/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java#L657).

The solution is to wrap the outermost `callInTransaction` in a `try`-`catch` block, `catch`ing `RuntimeException`, and doing the `isUniqueConstraintViolation` check there.

I am not sure why that behavior changed in this particular PR though...

sahibamittal:
> @sahibamittal The test is failing because this condition:
> 
> https://github.com/DependencyTrack/hyades-apiserver/blob/6f4ed40076eac3cd42c6ad172adfae30e13a969e/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java#L647-L652
> 
> can no longer happen in this location.
> 
> The unique constraint violation originating from the database will only be thrown when the DB transaction commits.
> 
> Because the statement in question is also wrapped in a `callInTransaction`:
> 
> https://github.com/DependencyTrack/hyades-apiserver/blob/6f4ed40076eac3cd42c6ad172adfae30e13a969e/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java#L607
> 
> the transaction will only commit after that call completes, so technically [here](https://github.com/DependencyTrack/hyades-apiserver/blob/6f4ed40076eac3cd42c6ad172adfae30e13a969e/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java#L657).
> 
> The solution is to wrap the outermost `callInTransaction` in a `try`-`catch` block, `catch`ing `RuntimeException`, and doing the `isUniqueConstraintViolation` check there.
> 
> I am not sure why that behavior changed in this particular PR though...

Thanks for checking Niklas 👍🏼 
I've moved `RuntimeException` outside but apart from `isUniqueConstraintViolation` it should throw exception as is else `ClientErrorException` being thrown inside will get caught here.

nscuro (approved):

PR #1046, opened 2025-01-31
Update FK constraints to enable cascading deletes

sahibamittal:
### Description

Recreate FK constraints to include the ON DELETE CASCADE clause.
Refactor the relevant application code of manual deletion logic.

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/1641

### Checklist

- [ ] I have read and understand the [contributing guidelines]

### DIFF
--- a/src/main/resources/migration/changelog-v5.6.0.xml
+++ b/src/main/resources/migration/changelog-v5.6.0.xml
@@ -189,4 +189,461 @@
             CREATE INDEX "COMPONENT_SHA3_512_IDX" ON "COMPONENT" ("SHA3_512") WHERE "SHA3_512" IS NOT NULL;
         </sql>
     </changeSet>
+
+    <changeSet id="v5.6.0-12" author="sahibamittal">
+        <dropForeignKeyConstraint baseTableName="AFFECTEDVERSIONATTRIBUTION" constraintName="AFFECTEDVERSIONATTRIBUTION_VULNERABILITY_FK"/>
+        <addForeignKeyConstraint baseColumnNames="VULNERABILITY" baseTableName="AFFECTEDVERSIONATTRIBUTION"
+             constraintName="AFFECTEDVERSIONATTRIBUTION_VULNERABILITY_FK" deferrable="true"
+             initiallyDeferred="true" onDelete="CASCADE" onUpdate="NO ACTION"
+             referencedColumnNames="ID" referencedTableName="VULNERABILITY" validate="true"/>
+
+        <dropForeignKeyConstraint baseTableName="AFFECTEDVERSIONATTRIBUTION" constraintName="AFFECTEDVERSIONATTRIBUTION_VULNERABLESOFTWARE_FK"/>
+        <addForeignKeyConstraint baseColumnNames="VULNERABLE_SOFTWARE" baseTableName="AFFECTEDVERSIONATTRIBUTION"
+                                 constraintName="AFFECTEDVERSIONATTRIBUTION_VULNERABLESOFTWARE_FK" deferrable="true"
+                                 initiallyDeferred="true" onDelete="CASCADE" onUpdate="NO ACTION"
+                                 referencedColumnNames="ID" referencedTableName="VULNERABLESOFTWARE" validate="true"/>
+
+        <dropForeignKeyConstraint baseTableName="ANALYSISCOMMENT" constraintName="ANALYSISCOMMENT_ANALYSIS_FK"/>
+        <addForeignKeyConstraint baseColumnNames="ANALYSIS_ID" baseTableName="ANALYSISCOMMENT"
+                                 constraintName="ANALYSISCOMMENT_ANALYSIS_FK" deferrable="true" initiallyDeferred="true"
+                                 onDelete="CASCADE" onUpdate="CASCADE" referencedColumnNames="ID"
+                                 referencedTableName="ANALYSIS" validate="true"/>
+
+        <dropForeignKeyConstraint baseTableName="ANALYSIS" constraintName="ANALYSIS_COMPONENT_FK"/>
+        <addForeignKeyConstraint baseColumnNames="COMPONENT_ID" baseTableName="ANALYSIS" constraintName="ANALYSIS_COMPONENT_FK"
+                                 deferrable="true" initiallyDeferred="true" onDelete="CASCADE" onUpdate="NO ACTION"
+                                 referencedColumnNames="ID" referencedTableName="COMPONENT" validate="true"/>
+
+        <dropForeignKeyConstraint baseTableName="ANALYSIS" constraintName="ANALYSIS_PROJECT_FK"/>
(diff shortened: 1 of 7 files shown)


src/main/resources/migration/changelog-v5.6.0.xml
```diff
@@ -189,4 +189,461 @@
...
+        <dropForeignKeyConstraint baseTableName="ANALYSIS" constraintName="ANALYSIS_VULNERABILITY_FK"/>
+        <addForeignKeyConstraint baseColumnNames="VULNERABILITY_ID" baseTableName="ANALYSIS"
+                                 constraintName="ANALYSIS_VULNERABILITY_FK" deferrable="true" initiallyDeferred="true"
+                                 onDelete="CASCADE" onUpdate="NO ACTION" referencedColumnNames="ID"
+                                 referencedTableName="VULNERABILITY" validate="true"/>
+
+        <dropForeignKeyConstraint baseTableName="ANALYSIS" constraintName="ANALYSIS_VULNERABILITY_POLICY_ID_FK"/>
+        <addForeignKeyConstraint baseTableName="ANALYSIS" baseColumnNames="VULNERABILITY_POLICY_ID"
+                                 constraintName="ANALYSIS_VULNERABILITY_POLICY_ID_FK" referencedTableName="VULNERABILITY_POLICY"
+                                 onDelete="CASCADE" referencedColumnNames="ID"/>
```
nscuro, line 231: This one should actually *not* use cascading deletes. An analysis should not be deleted when the vulnerability policy that was applied to it was removed, since that would also nuke the audit trail. There is a procedure here that resets the analysis and populates the audit trail accordingly: https://github.com/DependencyTrack/hyades-apiserver/blob/425cabc125664d75c856f26ee0b6ab8ea19ffda6/src/main/java/org/dependencytrack/persistence/jdbi/VulnerabilityPolicyDao.java#L211-L275

src/main/resources/migration/changelog-v5.6.0.xml
```diff
@@ -189,4 +189,461 @@
...
+
+        <dropForeignKeyConstraint baseTableName="COMPONENT" constraintName="COMPONENT_PROJECT_FK"/>
+        <addForeignKeyConstraint baseColumnNames="PROJECT_ID" baseTableName="COMPONENT" constraintName="COMPONENT_PROJECT_FK"
+                                 deferrable="true" initiallyDeferred="true" onDelete="CASCADE" onUpdate="NO ACTION"
+                                 referencedColumnNames="ID" referencedTableName="PROJECT" validate="true"/>
+
+        <dropForeignKeyConstraint baseTableName="COMPONENT" constraintName="COMPONENT_LICENSE_FK"/>
+        <addForeignKeyConstraint baseColumnNames="LICENSE_ID" baseTableName="COMPONENT" constraintName="COMPONENT_LICENSE_FK"
+                                 deferrable="true" initiallyDeferred="true" onDelete="CASCADE" onUpdate="NO ACTION"
+                                 referencedColumnNames="ID" referencedTableName="LICENSE" validate="true"/>
```
nscuro, line 276: This would cause the component to be deleted when the license is deleted, which is not what we would want. Better not use cascading deletes for this one.

src/main/resources/migration/changelog-v5.6.0.xml
```diff
@@ -189,4 +189,461 @@
...
+        <dropForeignKeyConstraint baseTableName="AFFECTEDVERSIONATTRIBUTION" constraintName="AFFECTEDVERSIONATTRIBUTION_VULNERABLESOFTWARE_FK"/>
+        <addForeignKeyConstraint baseColumnNames="VULNERABLE_SOFTWARE" baseTableName="AFFECTEDVERSIONATTRIBUTION"
+                                 constraintName="AFFECTEDVERSIONATTRIBUTION_VULNERABLESOFTWARE_FK" deferrable="true"
+                                 initiallyDeferred="true" onDelete="CASCADE" onUpdate="NO ACTION"
+                                 referencedColumnNames="ID" referencedTableName="VULNERABLESOFTWARE" validate="true"/>
+
+        <dropForeignKeyConstraint baseTableName="ANALYSISCOMMENT" constraintName="ANALYSISCOMMENT_ANALYSIS_FK"/>
+        <addForeignKeyConstraint baseColumnNames="ANALYSIS_ID" baseTableName="ANALYSISCOMMENT"
+                                 constraintName="ANALYSISCOMMENT_ANALYSIS_FK" deferrable="true" initiallyDeferred="true"
+                                 onDelete="CASCADE" onUpdate="CASCADE" referencedColumnNames="ID"
```
nscuro, line 209: Probably a typo?

```suggestion
                                 onDelete="CASCADE" onUpdate="NO ACTION" referencedColumnNames="ID"
```

src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
```diff
@@ -873,8 +875,10 @@ public Response deleteProject(
                 }
 
                 LOGGER.info("Project " + project + " deletion request by " + super.getPrincipal().getName());
-                try {
-                    qm.recursivelyDelete(project, true);
+
+                try (final Handle jdbiHandle = openJdbiHandle()) {
+                    final var projectDao = jdbiHandle.attach(ProjectDao.class);
+                    projectDao.deleteProject(project.getUuid());
```
nscuro, line 881: In cases like this where there's already a transaction ongoing (see `qm.runInTransaction` above), we must avoid opening another one. Problem being that each transaction allocates a database connection until it commits or is rolled back. Holding two connections at once is too wasteful.

Two options:

* Execute the deletion query using the `QueryManager` in scope (`qm`) instead.
* Create a JDBI instance using [`JdbiFactory#createLocalJdbi`](https://github.com/DependencyTrack/hyades-apiserver/blob/425cabc125664d75c856f26ee0b6ab8ea19ffda6/src/main/java/org/dependencytrack/persistence/jdbi/JdbiFactory.java#L138-L158) (you'd need to make that method `public` first).

nscuro (changes requested):

src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
```diff
@@ -873,8 +875,11 @@ public Response deleteProject(
                 }
 
                 LOGGER.info("Project " + project + " deletion request by " + super.getPrincipal().getName());
+
                 try {
-                    qm.recursivelyDelete(project, true);
+                    final Handle jdbiHandle = createLocalJdbi(qm).open();
```
nscuro, line 880: The `Handle` still needs closing to avoid resource leakage.

```suggestion
                try (final Handle jdbiHandle = createLocalJdbi(qm).open()) {
```

sahibamittal: Correct 👍🏼

nscuro (approved):

PR #1053, opened 2025-02-06
Issue 1600 : Add scheduled task for project retention maintenance

sahibamittal:
### Description

Add scheduled task for project retention maintenance

### Addressed Issue

https://github.com/DependencyTrack/hyades/issues/1600

Frontend: https://github.com/DependencyTrack/hyades-frontend/pull/236

Screenshot 2025-02-11 at 12 01 18

Disabled:
Screenshot 2025-02-11 at 16 02 57

### Checklist


src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
```diff
@@ -227,4 +228,39 @@ record ConciseProjectMetricsRow(
              WHERE "UUID" = :projectUuid
             """)
     int deleteProject(@Bind final UUID projectUuid);
+
+    @SqlUpdate("""
+           DELETE
+            FROM "PROJECT"
+            WHERE "PROJECT"."INACTIVE_SINCE" IS NOT NULL
+            AND NOW() - "PROJECT"."INACTIVE_SINCE" > :duration
+            AND "PROJECT"."ID" = any( array( SELECT "ID" FROM "PROJECT" LIMIT :batchSize));
```
nscuro, line 237: This condition may be problematic.

If none of the first `:batchSize` projects in the *entire* `PROJECT` table match the `INACTIVE_SINCE` conditions in the outer query, this query will never delete any project. In other words, the `LIMIT` should appear in the same query that filters on `INACTIVE_SINCE`.

I'd recommend to use a CTE here, like this:

```sql
WITH "CTE" AS (
  SELECT "ID"
    FROM "PROJECT"
   WHERE NOW() - "INACTIVE_SINCE" > :duration
   ORDER BY "INACTIVE_SINCE" -- Ensure oldest are deleted first.
   LIMIT :batchSize
)
DELETE
  FROM "PROJECT"
 WHERE "ID" IN (SELECT "ID" FROM "CTE")
```

An additional enhancement could be to calculate the "cutoff" timestamp ahead of time, so you can do:

```sql
-- Instant cutoff = Instant.now().minus(retentionDuration);
WHERE "INACTIVE_SINCE" < :cutoff
```

instead of:

```sql
WHERE NOW() - "INACTIVE_SINCE" > :duration
```

The benefit being that the former can utilize an index, whereas the latter can't. `NOW() - "INACTIVE_SINCE"` is a dynamic expression that needs to be evaluated for each row ad-hoc.

src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
```diff
@@ -227,4 +228,39 @@ record ConciseProjectMetricsRow(
...
+           DELETE
+            FROM "PROJECT"
+            WHERE "PROJECT"."INACTIVE_SINCE" IS NOT NULL
+            AND "PROJECT"."NAME" = :projectName
+            AND "PROJECT"."ID" NOT IN (
+                SELECT "PROJECT"."ID" 
+                 FROM "PROJECT"
+                 WHERE "PROJECT"."INACTIVE_SINCE" IS NOT NULL
+                 AND "PROJECT"."NAME" = :projectName
+                 ORDER BY "PROJECT"."INACTIVE_SINCE" DESC
```
nscuro, line 251: Won't `DESC` ordering not always delete the "youngest" inactive projects first?

src/main/java/org/dependencytrack/tasks/maintenance/ProjectMaintenanceTask.java
```diff
@@ -0,0 +1,114 @@
...
+    private static final Logger LOGGER = Logger.getLogger(ProjectMaintenanceTask.class);
+
+    @Override
+    public void inform(final Event event) {
+        if (!(event instanceof ProjectMaintenanceEvent)) {
+            return;
+        }
+
+        final long startTimeNs = System.nanoTime();
+        try (final Handle jdbiHandle = openJdbiHandle()) {
```
nscuro, line 53: This handle needs to be closed before `informLocked` is called, since it holds a DB connection while it remains open.

src/main/java/org/dependencytrack/tasks/maintenance/ProjectMaintenanceTask.java
```diff
@@ -0,0 +1,114 @@
...
+                            final var projectDao = jdbiHandle.attach(ProjectDao.class);
+                            return projectDao.deleteInactiveProjectsForRetentionDuration(retentionDuration, batchSize);
+                        });
+                numDeletedTotal.addAndGet(numDeletedLastBatch);
+            }
+        } else {
+            int retentionCadence = configPropertyDao.getValue(MAINTENANCE_INACTIVE_PROJECTS_RETENTION_CADENCE, Integer.class);
+            Integer projectLastBatch = null;
+            while (projectLastBatch == null || projectLastBatch > 0) {
+                projectLastBatch = withJdbiHandle(
```
nscuro, line 100: Since you're doing more than one query here, better do it in a transaction:

```suggestion
                projectLastBatch = inJdbiTransaction(
```

src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
sahibamittal: No, here sorting DESC will give list of youngest inactive projects and that list will be excluded : 
```
AND "PROJECT"."ID" NOT IN (
  SELECT "PROJECT"."ID" 
   FROM "PROJECT"
   WHERE "PROJECT"."INACTIVE_SINCE" IS NOT NULL
   AND "PROJECT"."NAME" = :projectName
   ORDER BY "PROJECT"."INACTIVE_SINCE" DESC
   LIMIT :retentionCadence
  )
```

src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java
sahibamittal: Nice catch, thanks for suggestion 👍🏼

src/main/java/org/dependencytrack/tasks/maintenance/ProjectMaintenanceTask.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    private record Statistics(int deletedInactiveProjects) {
+    }
+
+    private Statistics informLocked() {
+        assertLocked();
+        AtomicInteger numDeletedTotal = new AtomicInteger(0);
+
+        final Handle jdbiHandle = openJdbiHandle();
```
nscuro, line 81: This must be wrapped in a try-with-resources, or you should use `withJdbiHandle` instead (preferable). If anything fails between this, and `jdbiHandle.close()`, the backing DB connection would remain open.

nscuro: At the moment, the handle is also never closed when `MAINTENANCE_PROJECTS_RETENTION_ENABLE` is `false`, for example.

src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
```diff
@@ -45,6 +45,10 @@ public enum ConfigPropertyConstants {
     JIRA_USERNAME("integrations", "jira.username", null, PropertyType.STRING, "The optional username to authenticate with when creating an Jira issue", ConfigPropertyAccessMode.READ_WRITE),
     JIRA_PASSWORD("integrations", "jira.password", null, PropertyType.ENCRYPTEDSTRING, "The optional password for the username used for authentication", ConfigPropertyAccessMode.READ_WRITE),
     MAINTENANCE_METRICS_RETENTION_DAYS("maintenance", "metrics.retention.days", "90", PropertyType.INTEGER, "Number of days to retain metrics data for", ConfigPropertyAccessMode.READ_WRITE),
+    MAINTENANCE_PROJECTS_RETENTION_ENABLE("maintenance", "projects.retention.enable", "false", PropertyType.BOOLEAN, "Enable Inactive Project Deletion", ConfigPropertyAccessMode.READ_WRITE),
```
nscuro, line 48: Instead of introducing a separate flag, you could treat `MAINTENANCE_PROJECTS_RETENTION_TYPE != ('AGE', 'VERSIONS')` as disabled. So `null`, `foo`, `123`, etc. would all qualify as "not enabled". Defaulting the property to `null` disables the feature initially.

nscuro (approved):

PR #1072, opened 2025-02-28
Ingest component occurrences

nscuro:
### Description

Ingests component occurrences.

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/1691

### Additional Details

Frontend PR: https://github.com/DependencyTrack/hyades-frontend/pull/252

---

The [`additionalContext`](https://cyclonedx.org/docs/1.6/json/#components_items_evidence_occurrences_items_additionalContext) field is not considered for now. I have not yet seen any generator populating this field. Since its description is rather vague, I decided to opt for waiting until we know what kind of data to expect.

---

The primary key `ID` column of the new `COMPONENT_OCCURRENCE` table uses [UUIDv7](https://uuid7.com/). There are no separate `ID` and `UUID` column like in other legacy tables. This reduces storage requirements and overhead of maintaining two separate `UNIQUE` indexes.

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/jdbi/ApiRequestStatementCustomizer.java
+++ b/src/main/java/org/dependencytrack/persistence/jdbi/ApiRequestStatementCustomizer.java
@@ -67,11 +67,10 @@ import static org.jdbi.v3.core.generic.GenericTypes.parameterizeClass;
  *
  * @since 5.5.0
  */
-@SuppressWarnings("JavadocReference")
 class ApiRequestStatementCustomizer implements StatementCustomizer {
 
     static final String PARAMETER_PROJECT_ACL_TEAM_IDS = "projectAclTeamIds";
-    static final String TEMPLATE_PROJECT_ACL_CONDITION = "HAS_PROJECT_ACCESS(\"%s\".\"ID\", :projectAclTeamIds)";
+    static final String TEMPLATE_PROJECT_ACL_CONDITION = "HAS_PROJECT_ACCESS(%s, :projectAclTeamIds)";
 
     private final AlpineRequest apiRequest;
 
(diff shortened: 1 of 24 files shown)


src/main/java/org/dependencytrack/persistence/jdbi/ApiRequestStatementCustomizer.java
```diff
@@ -167,13 +166,11 @@ private void definePagination(final StatementContext ctx) {
...
-        if (apiRequest == null) {
-            return;
-        }
-
-        if (apiRequest.getPrincipal() == null
+        if (apiRequest == null
+            || apiRequest.getPrincipal() == null
             || !isAclEnabled(ctx)
             || hasAccessManagementPermission(ctx, apiRequest.getPrincipal())) {
+            ctx.define(ATTRIBUTE_API_PROJECT_ACL_CONDITION, "TRUE");
```
nscuro, line 173: We always define this variable now, this way our queries don't have to check if it's set etc.

src/main/java/org/dependencytrack/persistence/jdbi/DefineApiProjectAclCondition.java
```diff
@@ -48,12 +48,12 @@
...
      * @return Name of the attribute to define the condition as
      */
-    String name();
+    String name() default JdbiAttributes.ATTRIBUTE_API_PROJECT_ACL_CONDITION;
 
     /**
      * @return Alias of the {@code PROJECT} table to use in the condition
      */
-    String projectTableAlias();
+    String projectIdColumn();
```
nscuro, line 56: Changed this to accept a full column name, so the condition also works *without* having to join the `PROJECT` table first. See `ComponentDao#isAccessible` for example, where we can use the `PROJECT_ID` column directly.

sahibamittal (approved):
LGTM

PR #1081, opened 2025-03-04
Add risk rating vectors in findings response

sahibamittal:
### Description

Findings queries are missing CVSS vectors in their records.
Add the mapping for same.

### Addressed Issue

N/A

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/model/Finding.java
+++ b/src/main/java/org/dependencytrack/model/Finding.java
@@ -71,6 +71,8 @@ public class Finding implements Serializable {
                  , "VULNERABILITY"."SEVERITY"
                  , "VULNERABILITY"."CVSSV2BASESCORE"
                  , "VULNERABILITY"."CVSSV3BASESCORE"
+                 , "VULNERABILITY"."CVSSV2VECTOR"
+                 , "VULNERABILITY"."CVSSV3VECTOR"
                  , "VULNERABILITY"."OWASPRRLIKELIHOODSCORE"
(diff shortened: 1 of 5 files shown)


src/test/java/org/dependencytrack/resources/v1/FindingResourceTest.java
nscuro, line 1: Can we have a test / a few more assertions here that the vectors are indeed included in the API response(s)?

src/main/java/org/dependencytrack/model/Finding.java
```diff
@@ -71,6 +71,8 @@ public class Finding implements Serializable {
                  , "VULNERABILITY"."SEVERITY"
                  , "VULNERABILITY"."CVSSV2BASESCORE"
                  , "VULNERABILITY"."CVSSV3BASESCORE"
+                 , "VULNERABILITY"."CVSSV2VECTOR"
+                 , "VULNERABILITY"."CVSSV3VECTOR"
```
nscuro, line 75: Should also include `OWASPRRVECTOR` for completeness.

src/main/java/org/dependencytrack/model/Finding.java
```diff
@@ -119,6 +121,8 @@ public class Finding implements Serializable {
...
```
nscuro, line 125: Should also include `OWASPRRVECTOR` for completeness.

nscuro (changes requested):

nscuro (approved):

PR #1093, opened 2025-03-14
Prevent duplicates in `PROJECT_ACCESS_TEAMS` table

nscuro:
### Description

Prevents duplicates in `PROJECT_ACCESS_TEAMS` table.

### Addressed Issue

Fixes https://github.com/DependencyTrack/hyades/issues/1713

### Additional Details

N/A

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/model/Project.java
+++ b/src/main/java/org/dependencytrack/model/Project.java
@@ -35,14 +35,14 @@ import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.github.packageurl.MalformedPackageURLException;
 import com.github.packageurl.PackageURL;
 import io.swagger.v3.oas.annotations.media.Schema;
-import jakarta.validation.constraints.NotBlank;
-import jakarta.validation.constraints.NotNull;
-import jakarta.validation.constraints.Pattern;
-import jakarta.validation.constraints.Size;
 import org.dependencytrack.persistence.converter.OrganizationalContactsJsonConverter;
 import org.dependencytrack.persistence.converter.OrganizationalEntityJsonConverter;
 import org.dependencytrack.resources.v1.serializers.CustomPackageURLSerializer;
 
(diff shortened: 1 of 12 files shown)


src/main/java/org/dependencytrack/resources/v1/ProjectResource.java
```diff
@@ -555,7 +555,6 @@ public Response createProject(final Project jsonProject) {
                     LOGGER.error("Failed to create project %s".formatted(jsonProject), e);
                     throw new ServerErrorException(Response.Status.INTERNAL_SERVER_ERROR);
                 }
-                qm.updateNewProjectACL(project, principal);
```
nscuro, line 558: This call caused a unique constraint violation because access teams are already assigned earlier in the method. Essentially this call has been creating duplicate records before.

src/main/java/org/dependencytrack/model/Project.java
```diff
@@ -297,10 +298,9 @@ public enum FetchGroup {
     private Date inactiveSince;
 
     @Persistent(table = "PROJECT_ACCESS_TEAMS", defaultFetchGroup = "true")
-    @Join(column = "PROJECT_ID")
+    @Join(column = "PROJECT_ID", primaryKey = "PROJECT_ACCESS_TEAMS_PK")
     @Element(column = "TEAM_ID")
-    @Order(extensions = @Extension(vendorName = "datanucleus", key = "list-ordering", value = "name ASC"))
-    private List<Team> accessTeams;
+    private Set<Team> accessTeams;
```
nscuro, line 303: Making this a `Set` causes DataNucleus to generate a primary key on `PROJECT_ID` and `TEAM_ID`. The PK ensures that neither column can be `null`, and that rows are unique.

We mirror this behavior in the DB changelog, and create a primary key, just like DN would do it.

sahibamittal (approved):

PR #1106, opened 2025-03-24
Update findings `cwes` mapping

sahibamittal:
### Description

Findings `cwes` is supposed to be array of CWE objects. 

### Addressed Issue

N/A

### Checklist

### DIFF
--- a/src/main/resources/templates/findings/sarif.peb
+++ b/src/main/resources/templates/findings/sarif.peb
@@ -44,7 +44,12 @@
             "group": "{{ finding.component.group }}",
             "version": "{{ finding.component.version }}",
             "source": "{{ finding.vulnerability.source }}",
-            "cwes": "{{ finding.vulnerability.cwes }}",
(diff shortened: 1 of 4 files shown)


src/main/resources/templates/findings/sarif.peb
```diff
@@ -44,7 +44,12 @@
             "group": "{{ finding.component.group }}",
             "version": "{{ finding.component.version }}",
             "source": "{{ finding.vulnerability.source }}",
-            "cwes": "{{ finding.vulnerability.cwes }}",
+            "cwes": [{% for cwe in finding.vulnerability.cwes %}
+                {
+                  "cweId": "{{ cwe.cweId }}",
+                  "name": "{{ cwe.name }}"
+                }
+            ]{% if not loop.last %},{% endif %}{% endfor %},
```
nscuro, line 52: I think this should handle commas within the array so needs to be within the `[]`.

```suggestion
                }
                {% if not loop.last %},{% endif %}{% endfor %}
            ],
```

PR #1108, opened 2025-03-25
Refactor `Analysis` DN queries to JDBI

sahibamittal:
### Description

There are some queries written in DN for `Analysis` table, convert them to JDBI. 

### Addressed Issue

N/A

### Checklist


src/main/java/org/dependencytrack/resources/v1/AnalysisResource.java
```diff
@@ -179,37 +181,50 @@ public Response updateAnalysis(AnalysisRequest request) {
...
-
-            final String comment = StringUtils.trimToNull(request.getComment());
-            qm.makeAnalysisComment(analysis, comment, commenter);
-            analysis = qm.getAnalysis(component, vulnerability);
+            withJdbiHandle(handle -> handle.attach(AnalysisDao.class)
+                    .makeAnalysisComment(analysisId, StringUtils.trimToNull(request.getComment()), commenter));
+            qm.getPersistenceManager().refresh(analysisExisting);
+            var analysis = qm.getAnalysis(component, vulnerability);
             NotificationUtil.analyzeNotificationCriteria(qm, analysis, analysisStateChange, suppressionChange);
             return Response.ok(analysis).build();
```
nscuro, line 229: I think in this case it would make more sense to *not* use JDBI:

* The analysis is fetched via DN (`qm.getAnalysis`), so it's in a persistent state already.
* We're recreating JDBI handles multiple times throughout this method, each time re-attaching the DAO for small changes.
* Because JDBI bypasses DN, we now need to refresh the `Analysis` object with DN to reflect the changes we made.
* DN and JDBI won't share the same DB transaction.

It would be more pragmatic I think to just use the DataNucleus way throughout this method. OR convert it to JDBI completely.

src/main/java/org/dependencytrack/persistence/jdbi/AnalysisDao.java
```diff
@@ -33,4 +42,63 @@ public interface AnalysisDao {
...
     void createComments(@Bind List<Long> analysisId, @Bind String commenter, @Bind List<String> comment);
 
+    @SqlQuery("""
+            INSERT INTO "ANALYSISCOMMENT"
+              ("ANALYSIS_ID", "COMMENT", "COMMENTER", "TIMESTAMP")
+            VALUES
+              (:analysisId, :comment, :commenter, NOW())
+            RETURNING *
+            """)
+    @GetGeneratedKeys("*")
```
nscuro, line 52: `@GetGeneratedKeys` is only needed in conjunction with `@SqlUpdate`.

src/main/java/org/dependencytrack/persistence/jdbi/AnalysisDao.java
```diff
@@ -33,4 +42,63 @@ public interface AnalysisDao {
...
+               (:projectId, :componentId, :vulnId,
+               <#if state>
+                    :state,
+               <#else>
+                    'NOT_SET',
+               </#if>
+               :justification, :response, :details, :suppressed)
+            ON CONFLICT ("PROJECT_ID", "COMPONENT_ID", "VULNERABILITY_ID") DO UPDATE
+            SET
+               "SUPPRESSED" = :suppressed
```
nscuro, line 84: Will this not overwrite `SUPPRESSED` even if it wasn't supposed to be?

src/main/java/org/dependencytrack/persistence/jdbi/AnalysisDao.java
```diff
@@ -33,4 +42,63 @@ public interface AnalysisDao {
...
+    @SqlUpdate("""
+            INSERT INTO "ANALYSIS"
+               ("PROJECT_ID", "COMPONENT_ID", "VULNERABILITY_ID", "STATE", "JUSTIFICATION", "RESPONSE", "DETAILS", "SUPPRESSED")
+            VALUES
+               (:projectId, :componentId, :vulnId,
+               <#if state>
+                    :state,
+               <#else>
+                    'NOT_SET',
+               </#if>
```
nscuro, line 80: No need for templating here, you can just use [coalesce](https://neon.tech/postgresql/postgresql-tutorial/postgresql-coalesce).

```suggestion
               COALESCE(:state, 'NOT_SET'),
```

src/main/java/org/dependencytrack/util/AnalysisCommentUtil.java
```diff
@@ -22,55 +22,63 @@
...
 import org.dependencytrack.util.AnalysisCommentFormatter.AnalysisCommentField;
 
+import static org.dependencytrack.persistence.jdbi.JdbiFactory.withJdbiHandle;
 import static org.dependencytrack.util.AnalysisCommentFormatter.formatComment;
 
 public final class AnalysisCommentUtil {
 
     private AnalysisCommentUtil() { }
 
-    public static boolean makeStateComment(final QueryManager qm, final Analysis analysis, final AnalysisState analysisState, final String commenter) {
```
nscuro, line 34: If you'd like to have these methods use JDBI, it's best to accept a JDBI `Handle` as parameter. Otherwise, multiple invocations of these methods could never share a transaction.

The following should be possible:

```java
useJdbiTransaction(handle -> {
    AnalysisCommentUtil.makeStateComment(handle, ..., ...);
    AnalysisCommentUtil.makeJustificationComment(handle, ..., ...);
    // ...
});
```

In fact, you might want to consider moving all these methods to `AnalysisDao` instead.

nscuro (changes requested):

src/main/java/org/dependencytrack/resources/v1/AnalysisResource.java
sahibamittal: Makes sense 👍🏼 So, in order to convert it to JDBI completely, query should fetch in addition to `Analysis` all columns for tables `Vulnerability`, `Component`, `Project` in that case?

nscuro: > [...] query should fetch in addition to Analysis all columns for tables Vulnerability, Component, Project in that case?

No, I don't think those are needed in this context. We should avoid loading data that we don't need.

sahibamittal: So for analysis object being returned, we can manually set vulnerability and component to it?

nscuro: > [...] else they would be null.

But that is fine if there's no need to access component or vulnerability through the analysis object (i.e. `analysis.getComponent().getName()`).

We could use a separate `record` as return type here, which holds the component's and vulnerability's ID instead of the full objects. That would make it more obvious to the caller what to expect.

In general I think this is a great example of the [object-relational impedance mismatch](https://en.wikipedia.org/wiki/Object%E2%80%93relational_impedance_mismatch). In OOP / DataNucleus, you can do stuff like:

```java
analysis.getComponent().getVulnerabilities().getFirst().getVulnerableSoftware().getLast().getId()
```

This works because DN will just lazy-load all the relationships behind the scenes. That's convenient but also horribly inefficient. If you'd want to replicate this kind of object graph traversal with SQL, you'd more or less have to fetch the entire database content at once.

src/main/java/org/dependencytrack/persistence/jdbi/AnalysisDao.java
sahibamittal: It seems the `Boolean` is not getting checked below and it throws not-null constraint error if I try to send `:suppressed` as null.

So, I thought to keep `suppressed` as primitive boolean instead and evaluate it before calling the method.

```
INSERT INTO "ANALYSIS"
               ("PROJECT_ID", "COMPONENT_ID", "VULNERABILITY_ID", "STATE", "JUSTIFICATION", "RESPONSE", "DETAILS", "SUPPRESSED")
            VALUES
               (:projectId, :componentId, :vulnId,
               COALESCE(:state, 'NOT_SET'),
               :justification, :response, :details,
               COALESCE(:suppressed, false))
            ON CONFLICT ("PROJECT_ID", "COMPONENT_ID", "VULNERABILITY_ID") DO UPDATE
            SET
                <#if state>
                    "STATE" = :state
               </#if>
               <#if state>
                    , "SUPPRESSED" = :suppressed
               </#if>
               <#if justification>
                    , "JUSTIFICATION" = :justification
               </#if>
               <#if response>
                    , "RESPONSE" = :response
               </#if>
               <#if details>
                    , "DETAILS" = :details
               </#if>
            RETURNING *
```

nscuro: > It seems the Boolean is not getting checked below and it throws not-null constraint error if I try to send :suppressed as null.

Are you saying the snippet you posted *doesn't* work, or is the snippet the solution you used to work around the issue you described? The snippet does look solid to me, other than `<#if state>` being used to determine whether `SUPPRESSED` should be set.

src/main/java/org/dependencytrack/model/Analysis.java
```diff
@@ -76,22 +76,22 @@ public class Analysis implements Serializable {
     @Persistent(defaultFetchGroup = "true")
     @Column(name = "STATE", jdbcType = "VARCHAR", allowsNull = "false")
     @NotNull
-    private AnalysisState analysisState;
+    private AnalysisState state;
```
nscuro, line 79: I understand the itch to rename these fields, but in doing so we're introducing a breaking API change (because fields in REST API responses change, too), and clients including the frontend will need to be updated. I'd prefer to avoid this if possible.

src/main/java/org/dependencytrack/persistence/jdbi/AnalysisDao.java
```diff
@@ -59,7 +61,17 @@ default AnalysisComment makeAnalysisComment(long analysisId, String comment, Str
...
 
-    @SqlUpdate("""
+    @SqlQuery("""
+            SELECT *
+            FROM "ANALYSIS"
+            WHERE "COMPONENT_ID" = :componentId
+            AND "VULNERABILITY_ID" = :vulnId
+            LIMIT 1
+            """)
+    @RegisterBeanMapper(Analysis.class)
```
nscuro, line 71: It seems you can simply replace this with `@RegisterRowMapper(AnalysisRowMapper.class)` to solve your original issue of some fields not getting mapped correctly?

src/main/java/org/dependencytrack/model/Analysis.java
sahibamittal: Right, I'll update the fields in query instead 👍🏼

src/main/java/org/dependencytrack/resources/v1/AnalysisResource.java
```diff
@@ -179,40 +181,48 @@ public Response updateAnalysis(AnalysisRequest request) {
...
-                suppressionChange = AnalysisCommentUtil.makeAnalysisSuppressionComment(qm, analysis, request.isSuppressed(), commenter);
-                analysis = qm.makeAnalysis(component, vulnerability, request.getAnalysisState(), request.getAnalysisJustification(), request.getAnalysisResponse(), request.getAnalysisDetails(), request.isSuppressed());
             } else {
-                analysis = qm.makeAnalysis(component, vulnerability, request.getAnalysisState(), request.getAnalysisJustification(), request.getAnalysisResponse(), request.getAnalysisDetails(), request.isSuppressed());
-                analysisStateChange = true; // this is a new analysis - so set to true because it was previously null
-                if (AnalysisState.NOT_SET != request.getAnalysisState()) {
-                    qm.makeAnalysisComment(analysis, formatComment(AnalysisCommentField.STATE, AnalysisState.NOT_SET, request.getAnalysisState()), commenter);
-                }
+                commenter = null;
             }
```
nscuro, line 195: This method is still a mixture of DataNucleus and JDBI:

* Project, component, and vulnerability are fetched via DataNucleus.
* The analysis is modified via JDBI.
* The modified analysis is fetched again via DataNucleus and passed to `NotificationUtil.analyzeNotificationCriteria`.

Trying to mix these two here feels like bending over backward. Can you look into using only one or the other?

For example, [`NotificationUtil#analyzeNotificationCritera`](https://github.com/DependencyTrack/hyades-apiserver/blob/f240a2c4080ab190653d703127c6d38379b7c365/src/main/java/org/dependencytrack/util/NotificationUtil.java#L84-L90) is only used in this method. I assume you're having to fetch the `Analysis` object from DataNucleus again to make that method call work. Another option you could check to avoid this is to use [`NotificationSubjectDao#getForProjectAuditChange`](https://github.com/DependencyTrack/hyades-apiserver/blob/f240a2c4080ab190653d703127c6d38379b7c365/src/main/java/org/dependencytrack/persistence/jdbi/NotificationSubjectDao.java#L243-L330), which is what [we already use during vuln scan result processing](https://github.com/DependencyTrack/hyades-apiserver/blob/f240a2c4080ab190653d703127c6d38379b7c365/src/main/java/org/dependencytrack/event/kafka/processor/VulnerabilityScanResultProcessor.java#L694-L705).

The `requireAccess` check at the top you could replace with [`requireComponentAccess`](https://github.com/DependencyTrack/hyades-apiserver/blob/f240a2c4080ab190653d703127c6d38379b7c365/src/main/java/org/dependencytrack/resources/v1/AbstractApiResource.java#L83-L105), which will also take care of checking whether the component (and implicitly the project) even exists.

Ideally you'd be able to remove `QueryManager` entirely from this method.

sahibamittal: And convert fetch methods into JDBI for `project`, `component` and `vulnerability`?
Instead of fetching whole objects, I'll fetch their IDs only required further for `Analysis` JDBIs. Or instead of fetching data, I can simply add joins with respective tables in `Analysis` queries.

nscuro (approved):

PR #1111, opened 2025-03-28
Add pagination to `/api/v1/finding` endpoints

sahibamittal:
### Description

Refactor the findings endpoints to support JDBI pagination.
1. `/v1/finding/project/{uuid}`
2. `/v1/finding`
3. `/v1/finding/grouped`

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/1708

### Checklist

### DIFF
--- a/src/main/java/org/dependencytrack/persistence/jdbi/FindingDao.java
+++ b/src/main/java/org/dependencytrack/persistence/jdbi/FindingDao.java
@@ -18,21 +18,13 @@
  */
 package org.dependencytrack.persistence.jdbi;
 
-import alpine.persistence.PaginatedResult;
-import alpine.resources.AlpineRequest;
 import alpine.server.util.DbUtil;
-import com.github.packageurl.PackageURL;
 import org.dependencytrack.model.AnalysisState;
 import org.dependencytrack.model.AnalyzerIdentity;
 import org.dependencytrack.model.Finding;
-import org.dependencytrack.model.GroupedFinding;
-import org.dependencytrack.model.RepositoryMetaComponent;
-import org.dependencytrack.model.RepositoryType;
 import org.dependencytrack.model.Severity;
 import org.dependencytrack.model.Vulnerability;
 import org.dependencytrack.model.VulnerabilityAlias;
-import org.dependencytrack.persistence.RepositoryQueryManager.RepositoryMetaComponentSearch;
-import org.dependencytrack.util.PurlUtil;
 import org.jdbi.v3.json.Json;
 import org.jdbi.v3.sqlobject.config.RegisterConstructorMapper;
 import org.jdbi.v3.sqlobject.customizer.AllowUnusedBindings;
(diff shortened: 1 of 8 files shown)


nscuro:
I should've added this to the original issue, but the problem with the existing pagination for findings search is that it happens in-memory, in the application: https://github.com/DependencyTrack/dependency-track/issues/4094

At that point it already loaded all the data from the database. In order for pagination to provide actual benefit, it needs to happen at the database-level.

For JDBI, you can achieve this by adding the following to the query methods:

https://github.com/DependencyTrack/hyades-apiserver/blob/f44550f919ac091de69524c7b4b33cfcb74a0d8e/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java#L155

Which then gets expanded to something like this:

```sql
OFFSET 0 FETCH NEXT 100 ROWS ONLY
```

This will need to be combined with the usual:

https://github.com/DependencyTrack/hyades-apiserver/blob/f44550f919ac091de69524c7b4b33cfcb74a0d8e/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java#L83

to compute the total number of items across all pages.

src/main/java/org/dependencytrack/persistence/jdbi/FindingDao.java
```diff
@@ -164,13 +158,13 @@ record GroupedFindingRow(
                 ON "COMPONENT"."PROJECT_ID" = "PROJECT"."ID"
              WHERE "COMPONENT"."PROJECT_ID" = :projectId
                AND (:includeSuppressed OR "ANALYSIS"."SUPPRESSED" IS NULL OR NOT "ANALYSIS"."SUPPRESSED")
```
nscuro, line 160: One thing we now need to take care of is ordering. If we we don't specify any `ORDER BY` clause, Postgres does not guarantee the order in which records are returned, which of course is problematic for clients iterating over multiple pages.

For every query, we need a column with unique values, that we can always order by to ensure stable ordering. You can use `@AllowApiOrdering(alwaysBy = "columnName")` for that: https://github.com/DependencyTrack/hyades-apiserver/blob/5832aedf8c99b1dd918d4fe2dfc51ba0effc58a5/src/main/java/org/dependencytrack/persistence/jdbi/AllowApiOrdering.java#L50-L56

For example: https://github.com/DependencyTrack/hyades-apiserver/blob/5832aedf8c99b1dd918d4fe2dfc51ba0effc58a5/src/main/java/org/dependencytrack/persistence/jdbi/ProjectDao.java#L162-L165

Since a finding is a result of multiple joins and doesn't technically have an ID on its own, I'd suggest we always order by `FINDINGATTRIBUTION.ID`. Ordering by vulnerability ID or similar would not be sufficient b/c it isn't necessarily unique within a result set.

nscuro (approved):
Great work!

PR #1121, opened 2025-04-07
Convert to multi-module Maven project (phase 1)

nscuro:
### Description

Converts to multi-module Maven project.

This implements phase 1 of https://github.com/DependencyTrack/hyades/issues/1710, by merely shifting the previous code to a new `apiserver` module.

The main module has been renamed to `dependency-track-parent`. It currently only holds a very small set of common configurations.

Only the `apiserver` module pulls in Alpine. This enables future modules to be completely independent of Alpine.

### Addressed Issue

Relates to https://github.com/DependencyTrack/hyades/issues/1710

### Additional Details

Hyades PR: https://github.com/DependencyTrack/hyades/pull/1749

---

Noteworthy changes:

1. We can no longer [inherit from Alpine](https://github.com/DependencyTrack/hyades-apiserver/blob/36c8d519e83ec1e5a90a620fdad508ea451effb5/pom.xml#L24-L28) as parent project. Maven modules can only have one parent. Importing another POM like [so](https://github.com/DependencyTrack/hyades/blob/db23e624d7d67ac52b8ceb22c082812164a00e40/pom.xml#L117-L123) only imports `dependencies`, but not `properties` or `plugins`.
    * Hence, plugins previously inherited from Alpine are now directly included in `apiserver/pom.xml`.
2. The Maven profiles `enhance` and `embedded-jetty` are no longer necessary. The `apiserver` module embeds the respective `plugin` definitions without profiles, since it always needs them anyway.
3. The `Release CI` workflow had to be modified to push file changes via `git` command rather than GitHub CLI. The latter only supported committing of individual files (i.e. *the* `pom.xml`). Since we now have multiple `pom.xml`s, that did no longer cut it.

### Checklist

- [x] This PR introduces new or alters existing behavior, and I have updated the [documentation] accordingly


nscuro:
buf's breaking changes check is failing because the directory `apiserver/src/main/proto` doesn't exist in `main` yet. Checks should continue to function once this PR is merged.

sahibamittal (approved):

PR #1123, opened 2025-04-08
Allow reuse of project name and version when different parent project

jhoward-lm:
### Description

This PR modifies the existing unique constraint on a project's combined name and version columns to include the parent project ID as an additional factor defining uniqueness.

The goal is to allow reuse of a project name, or project name and version, when the child projects have different parent projects.

```mermaid
---
title: New Behavior
---
erDiagram
    p1["parent project"] {
        NAME parent-project
        VERSION v1
    }

    p2["parent project"] {
        NAME parent-project
        VERSION v2
    }

    c1["child project"] {
        NAME child-project
        VERSION v1
    }

    c2["child project"] {
        NAME child-project
        VERSION v2
    }

    c3["child project"] {
        NAME child-project
        VERSION v1
    }

    c4["child project"] {
        NAME child-project
        VERSION v2
    }

    p1 |o--o{ c1 : "parent of"
    p1 |o--o{ c2 : "parent of"
    p2 |o--o{ c3 : "parent of"
    p2 |o--o{ c4 : "parent of"
```

### Checklist

- [X] This PR introduces changes to the database model, and I have updated the [migration changelog] accordingly

### DIFF
--- a/src/main/resources/migration/changelog-v5.6.0.xml
+++ b/src/main/resources/migration/changelog-v5.6.0.xml
@@ -1349,4 +1349,38 @@
             ALTER TABLE "VULNERABILITY" ALTER COLUMN "SEVERITY" TYPE severity USING "SEVERITY"::severity;
         </sql>
     </changeSet>
+
+    <changeSet id="v5.6.0-19" author="jhoward-lm">
+        <dropIndex tableName="PROJECT" indexName="PROJECT_NAME_VERSION_IDX" />
+        <dropIndex tableName="PROJECT" indexName="PROJECT_NAME_VERSION_NULL_IDX" />
+
+        <sql splitStatements="true">
+            -- Remove all duplicate projects that would violate the unique indices before creating
+            DELETE FROM "PROJECT"
+             WHERE "ID" NOT IN (
+               SELECT MIN("ID")
+                 FROM "PROJECT"
+                GROUP BY 
+                    "NAME", 
+                    COALESCE("PARENT_PROJECT_ID", -1), 
+                    COALESCE("VERSION", 'NULL_PLACEHOLDER')
+            );
+
+            -- Unique index that excludes rows where PARENT_PROJECT_ID or VERSION is NULL
+            CREATE UNIQUE INDEX "PROJECT_NAME_VERSION_PARENT_IDX"
+                ON "PROJECT" ("NAME", "PARENT_PROJECT_ID", "VERSION")
+             WHERE "PARENT_PROJECT_ID" IS NOT NULL
+               AND "VERSION" IS NOT NULL;
+
+            -- Unique index to handle cases where PARENT_PROJECT_ID is NULL
+            CREATE UNIQUE INDEX "PROJECT_NAME_VERSION_NULL_PARENT_IDX"
+                ON "PROJECT" ("NAME", "VERSION")
+             WHERE "PARENT_PROJECT_ID" IS NULL;
+
+            -- Unique index to handle cases where VERSION is NULL
+            CREATE UNIQUE INDEX "PROJECT_NAME_PARENT_NULL_VERSION_IDX"
+                ON "PROJECT" ("NAME", "PARENT_PROJECT_ID")
+             WHERE "VERSION" IS NULL;
+        </sql>
+    </changeSet>
 </databaseChangeLog>


src/main/resources/migration/changelog-v5.6.0.xml
```diff
@@ -1349,4 +1349,38 @@
...
+            -- Remove all duplicate projects that would violate the unique indices before creating
+            DELETE FROM "PROJECT"
+             WHERE "ID" NOT IN (
+               SELECT MIN("ID")
+                 FROM "PROJECT"
+                GROUP BY 
+                    "NAME", 
+                    COALESCE("PARENT_PROJECT_ID", -1), 
+                    COALESCE("VERSION", 'NULL_PLACEHOLDER')
+            );
```
nscuro, line 1367: Deleting projects is a rather destructive action, I think it would be better to perhaps `UPDATE` the version of duplicates instead. For example via:

```sql
SET "VERSION" = format('%s %s', "VERSION", "UUID") 
```

This query modifying any rows at all should be pretty rare at this stage, but still, let's not delete projects without user consent.

src/main/resources/migration/changelog-v5.6.0.xml
```diff
@@ -1349,4 +1349,38 @@
...
+            -- Unique index that excludes rows where PARENT_PROJECT_ID or VERSION is NULL
+            CREATE UNIQUE INDEX "PROJECT_NAME_VERSION_PARENT_IDX"
+                ON "PROJECT" ("NAME", "PARENT_PROJECT_ID", "VERSION")
+             WHERE "PARENT_PROJECT_ID" IS NOT NULL
+               AND "VERSION" IS NOT NULL;
+
+            -- Unique index to handle cases where PARENT_PROJECT_ID is NULL
+            CREATE UNIQUE INDEX "PROJECT_NAME_VERSION_NULL_PARENT_IDX"
+                ON "PROJECT" ("NAME", "VERSION")
+             WHERE "PARENT_PROJECT_ID" IS NULL;
```
nscuro, line 1378:
```suggestion
             WHERE "PARENT_PROJECT_ID" IS NULL
               AND "VERSION" IS NOT NULL;
```

src/main/resources/migration/changelog-v5.6.0.xml
```diff
@@ -1349,4 +1349,38 @@
...
+
+            -- Unique index to handle cases where VERSION is NULL
+            CREATE UNIQUE INDEX "PROJECT_NAME_PARENT_NULL_VERSION_IDX"
+                ON "PROJECT" ("NAME", "PARENT_PROJECT_ID")
+             WHERE "VERSION" IS NULL;
```
nscuro, line 1383:
```suggestion
             WHERE "VERSION" IS NULL
               AND "PARENT_PROJECT_ID" IS NOT NULL;
```

nscuro (changes requested):

src/main/resources/migration/changelog-v5.6.0.xml
jhoward-lm: I changed the `DELETE` to

```sql
UPDATE "PROJECT"
   SET "VERSION" = format('%s-%s', "VERSION", split_part("UUID"::TEXT, '-', 1))
```

which results in version strings like `v0.1.0-7ebdbfb8`. Are you ok with that?

nscuro:
`createProjectWithoutVersionDuplicateTest` is failing because we don't detect duplicates when both `VERSION` and `PARENT_PROJECT_ID` are null.

I think an additional index is needed:

```sql
CREATE UNIQUE INDEX "PROJECT_NAME_NULL_PARENT_NULL_VERSION_IDX"
    ON "PROJECT" ("NAME")
 WHERE "VERSION" IS NULL
   AND "PARENT_PROJECT_ID" IS NULL;
```

nscuro:
> @jhoward-lm Do you think there are any changes to the models that will be needed to support this?

Upon initial look, there is this method which checks whether a given name-version combination already exists:

https://github.com/DependencyTrack/hyades-apiserver/blob/c1f012391448203f16ae425a9911df68ce15ae33/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java#L1256-L1288

This would need updating to take the parent project into consideration.

We also have this `lookup` method for projects which, given a name and version, can only return a single result:

https://github.com/DependencyTrack/hyades-apiserver/blob/c1f012391448203f16ae425a9911df68ce15ae33/src/main/java/org/dependencytrack/resources/v1/ProjectResource.java#L303-L339

The logic to retrieve all versions of a project is also solely based on the project name at the moment:

https://github.com/DependencyTrack/hyades-apiserver/blob/c1f012391448203f16ae425a9911df68ce15ae33/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java#L1327-L1333

There are multiple usages of this method, which retrieves a single project by name and version:

https://github.com/DependencyTrack/hyades-apiserver/blob/c1f012391448203f16ae425a9911df68ce15ae33/src/main/java/org/dependencytrack/persistence/ProjectQueryManager.java#L251-L280

jhoward-lm:
> `createProjectWithoutVersionDuplicateTest` is failing because we don't detect duplicates when both `VERSION` and `PARENT_PROJECT_ID` are null.
> 
> I think an additional index is needed:
> 
> ```sql
> CREATE UNIQUE INDEX "PROJECT_NAME_NULL_PARENT_NULL_VERSION_IDX"
>     ON "PROJECT" ("NAME")
>  WHERE "VERSION" IS NULL
>    AND "PARENT_PROJECT_ID" IS NULL;
> ```

You're right, I had already created it and testing locally now.

Also, do the existing non-unique indexes `PROJECT_NAME_IDX` and `PROJECT_VERSION_IDX` need to stay, or is their functionality covered by the new unique indexes on those columns?

nscuro:
> Also, do the existing non-unique indexes PROJECT_NAME_IDX and PROJECT_VERSION_IDX need to stay, or is their functionality covered by the new unique indexes on those columns?

They'd be covered with these new indexes so are safe to drop.

jhoward-lm:
Going to defer this one until after resolution of #1121 and DependencyTrack/hyades#1736/DependencyTrack/hyades#1759

PR #1124, opened 2025-04-10
Ensure tag name and relationship uniqueness

nscuro:
### Description

Ensures tag name and relationship uniqueness.

* Adds missing (unique) index on tag names.
* Adds composite primary keys to all join tables of tag relationships. Thus prevents duplicate entries, and voids the need for indexes on individual columns.
* Changes tag relationships from ordered `List`s to `Set`s. We don't need ordering here, and it only adds overhead.
* Changes `equals` and `hashCode` of `Tag` to be based on `name`, rather than `id`.
* Removes and re-wires any duplicate records during the DB migration.

### Addressed Issue

Fixes https://github.com/DependencyTrack/hyades/issues/1755

### Additional Details

N/A

### Checklist

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/integrations/kenna/KennaDataTransformer.java
+++ b/apiserver/src/main/java/org/dependencytrack/integrations/kenna/KennaDataTransformer.java
@@ -36,6 +36,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import static org.dependencytrack.persistence.jdbi.JdbiFactory.withJdbiHandle;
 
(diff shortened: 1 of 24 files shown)


sahibamittal (dismissed):
LGTM 👍🏼

nscuro:
@sahibamittal Had to rebase this PR to resolve a merge conflict caused by #1126. Please have another look when you get the chance!

sahibamittal (approved):

src/main/resources/migration/changelog-v5.6.0.xml
```diff
@@ -1398,4 +1398,124 @@
...
         </createIndex>
     </changeSet>
+
+    <changeSet id="v5.6.0-20" author="nscuro">
+        <!--
+          Disable deferred constraint checking during this transaction.
+          Without this, we would not be able to add primary keys after
+          having deleted records from a table before.
+        -->
+        <sql>SET CONSTRAINTS ALL IMMEDIATE</sql>
```
nscuro, line 1408: I did a final test run with duplicate records across all tag tables, and the migration ended up failing with:

```
cannot ALTER TABLE "NOTIFICATIONRULE_TAGS" because it has pending trigger events
```

This happened when the migration tried to add primary keys. It turns out that, if a table has deferred foreign key constraints, you cannot modify rows and alter the table schema in the same transaction: https://github.com/friek/til/blob/master/postgres/alter-table-pending-triggers.md

The solution is to disable deferred constraints during the transaction, such that foreign keys are checked immediately and cannot block table altering.

sahibamittal (approved):

sahibamittal (approved):

PR #1130, opened 2025-04-16
Convert to multi-module Maven project (phase 2.1)

nscuro:
### Description

Moves Protobuf files into separate Maven module. 

This way, code generation and compilation of generated source files does not have to run whenever API server code needs compiling. Analogue to the `proto` module in the hyades repository: https://github.com/DependencyTrack/hyades/tree/main/proto

Also introduces a new `coverage-report` module which is required to aggregate test coverage from multiple modules, as outlined here: https://github.com/jacoco/jacoco/wiki/MavenMultiModule

### Addressed Issue

Relates to https://github.com/DependencyTrack/hyades/issues/1710

### Checklist


apiserver/src/test/java/org/dependencytrack/util/ConfigUtilTest.java
```diff
@@ -34,6 +35,7 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+@Ignore // Dirties the JVM which doesn't play nicely with Maven Surefire reusing a JVM for all tests.
```
nscuro, line 38: Necessary to make use of `reuseForks=true` in Maven. This test being disabled is a fair tradeoff for tests running significantly faster IMO.

apiserver/pom.xml
```diff
@@ -735,6 +702,7 @@
                             <mainClass>alpine.embedded.EmbeddedJettyServer</mainClass>
                         </manifest>
                     </archive>
+                    <attachClasses>true</attachClasses>
```
nscuro, line 705: [Necessary](https://maven.apache.org/plugins/maven-war-plugin/war-mojo.html#attachClasses) for the `apiserver` to be usable as dependency in other modules (i.e. `coverage-report`). Normally, modules with `packaging=war`can't be used as dependency.

nscuro:
Buf is again expected to fail because the location of proto files has changed. It should start working again once the PR is merged.

sahibamittal (approved):

PR #1141, opened 2025-04-24
Partition metrics tables

sahibamittal:
### Description

Introducing partitions on metrics tables.

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/1744

### Checklist


apiserver/src/main/resources/migration/procedures/procedure_migrate_metrics_to_partitions.sql
```diff
@@ -0,0 +1,53 @@
...
+    next_day DATE;
+BEGIN
+    -- Fetch retention value from config
+    SELECT COALESCE(cp."PROPERTYVALUE"::INTEGER, 90)
+    INTO retention_days
+    FROM "CONFIGPROPERTY" cp
+    WHERE cp."GROUPNAME" = 'maintenance'
+      AND cp."PROPERTYNAME" = 'metrics.retention.days';
+
+    start_date := current_date - retention_days;
```
nscuro, line 23: Nit: Let's make this more explicit. I had to check first if subtracting an integer from a date even has the intended effect 😅 

```suggestion
    start_date := current_date - (retention_days * '1 day'::interval);
```

apiserver/src/main/resources/migration/procedures/procedure_migrate_metrics_to_partitions.sql
```diff
@@ -0,0 +1,53 @@
...
+    target_table TEXT,
+    source_table TEXT
+)
+LANGUAGE "plpgsql"
+AS
+$$
+DECLARE
+    retention_days INTEGER := 90;
+    start_date DATE;
+    end_date DATE := current_date;
```
nscuro, line 11: We could take this opportunity to create the next partition already:

```suggestion
    end_date DATE := current_date + '1 day'::interval;
```

apiserver/src/main/resources/migration/changelog-v5.6.0.xml
```diff
@@ -1430,4 +1430,201 @@
...
+
+            -- drop old table
+            DROP TABLE "DEPENDENCYMETRICS";
+
+            -- rename partitioned table
+            ALTER TABLE "DEPENDENCYMETRICS_NEW" RENAME TO "DEPENDENCYMETRICS";
+
+            -- rename PK
+            ALTER TABLE "DEPENDENCYMETRICS" RENAME CONSTRAINT "DEPENDENCYMETRICS_NEW_PK" TO "DEPENDENCYMETRICS_PK";
+
```
nscuro, line 1611: The `ID` column creates a sequence behind the scenes, which we'll need to rename as well:

```sql
ALTER SEQUENCE "DEPENDENCYMETRICS_NEW_ID_seq" RENAME TO "DEPENDENCYMETRICS_ID_seq";
```

apiserver/src/main/resources/migration/procedures/procedure_migrate_metrics_to_partitions.sql
nscuro, line 1: Since this is a short-lived procedure, it should be embedded into `changelog-v5.6.0.xml` directly. Otherwise, if we remove or modify this file later, it would corrupt the changelog.

apiserver/src/main/resources/migration/changelog-v5.6.0.xml
```diff
@@ -1430,4 +1430,201 @@
...
+                "POLICYVIOLATIONS_WARN" int4 NULL,
+                "PROJECT_ID" int8 NOT NULL,
+                "SUPPRESSED" int4 NOT NULL,
+                "UNASSIGNED_SEVERITY" int4 NULL,
+                "VULNERABILITIES" int4 NOT NULL,
+                CONSTRAINT "DEPENDENCYMETRICS_NEW_PK" PRIMARY KEY ("ID", "LAST_OCCURRENCE")
+            ) PARTITION BY RANGE ("LAST_OCCURRENCE");
+
+            -- create new indexes
+            CREATE INDEX "DEPENDENCYMETRICS_NEW_FIRST_OCCURRENCE_IDX" ON "DEPENDENCYMETRICS_NEW" ("FIRST_OCCURRENCE");
```
nscuro, line 1597: With records being append-only, the distinction between first and last occurrence will have no meaning anymore. An index on `FIRST_OCCURRENCE` is no longer necessary.

We should rename `LAST_OCCURRENCE` to `TIMESTAMP` or something later.

apiserver/src/main/java/org/dependencytrack/tasks/metrics/MetricsPartitionCreateTask.java
```diff
@@ -0,0 +1,62 @@
...
+    private void createPartitions() {
+        LOGGER.info("Executing creation of partitions for Portfolio, Project and Dependency metrics");
+        final long startTimeNs = System.nanoTime();
+        try {
+            Metrics.createMetricsPartitions();
+        } finally {
+            LOGGER.info("Completed creating metrics partitions in " + Duration.ofNanos(System.nanoTime() - startTimeNs));
+        }
+    }
+}
```
nscuro, line 62: Would it make sense to incorporate this into the existing `MetricsMaintenanceTask`

apiserver/src/main/resources/migration/procedures/procedure_migrate_metrics_to_partitions.sql
```diff
@@ -0,0 +1,53 @@
...
+BEGIN
+    -- Fetch retention value from config
+    SELECT COALESCE(cp."PROPERTYVALUE"::INTEGER, 90)
+    INTO retention_days
+    FROM "CONFIGPROPERTY" cp
+    WHERE cp."GROUPNAME" = 'maintenance'
+      AND cp."PROPERTYNAME" = 'metrics.retention.days';
+
+    start_date := current_date - retention_days;
+    partition_date := start_date;
```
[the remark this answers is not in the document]
sahibamittal: Great suggestion!

apiserver/src/main/java/org/dependencytrack/persistence/MetricsQueryManager.java
nscuro, line 1: Can you do another pass on methods in here to see which of them are no longer needed? 

Did some high-level checks and I think all methods that operate on `ProjectMetrics`, `DependencyMetrics` and `PortfolioMetrics` should be removable.

apiserver/src/main/resources/migration/changelog-v5.6.0.xml
```diff
@@ -1430,4 +1430,282 @@
...
+                "POLICYVIOLATIONS_SECURITY_UNAUDITED" int4 NULL,
+                "POLICYVIOLATIONS_TOTAL" int4 NULL,
+                "POLICYVIOLATIONS_UNAUDITED" int4 NULL,
+                "POLICYVIOLATIONS_WARN" int4 NULL,
+                "PROJECT_ID" int8 NOT NULL,
+                "SUPPRESSED" int4 NOT NULL,
+                "UNASSIGNED_SEVERITY" int4 NULL,
+                "VULNERABILITIES" int4 NOT NULL,
+                "VULNERABLECOMPONENTS" int4 NOT NULL,
+                CONSTRAINT "PROJECTMETRICS_NEW_PK" PRIMARY KEY ("ID", "LAST_OCCURRENCE")
```
nscuro, line 1617: We may want to use `PROJECT_ID, LAST_OCCURRENCE` as primary key instead:

* Our lookups are almost always `where project_id = ? and last_occurrence >= ?`, so we'd need an index on both columns anyway. Primary keys create a unique index behind the scenes already.
* We want to enforce uniqueness such that there can be no duplicate records for the same `PROJECT_ID`<->`LAST_OCCURRENCE` combination. A primary key achieves that.

Following this, the `ID` column would be no longer needed.

For `PORTFOLIOMETRICS`, `LAST_OCCURRENCE` alone can form a primary key.

apiserver/src/main/resources/migration/changelog-v5.6.0.xml
```diff
@@ -1430,4 +1430,282 @@
...
+                    );
+
+                    partition_date := next_day;
+                END LOOP;
+            END;
+            $$;
+            ]]>
+        </sql>
+        <!--Migration for "PORTFOLIOMETRICS"-->
+        <sql>
```
nscuro, line 1521: We will need to split the data migrations up into multiple `changeSet`s. Since every change set runs in a transaction, trying to migrate all the data in one is very likely to cause issues.

I would propose the following order:

* Changeset `v5.6.0-23.1`
  1. Create `MIGRATE_METRICS_TO_PARTITIONS` procedure
  2. Create `PORTFOLIOMETRICS_NEW` table
  3. Create `PROJECTMETRICS_NEW` table
  4. Create `DEPENDENCYMETRICS_NEW` table
* Changeset `v5.6.0-23.2`
  1. Migrate `PORTFOLIOMETRICS` data
  2. Drop old `PORTFOLIOMETRICS` table
  3. Rename table and index
* Changeset `v5.6.0-23.3`
  1. Migrate `PROJECTMETRICS` data
  2. Drop old `PROJECTOMETRICS` table
  3. Rename table and index
* Changeset `v5.6.0-23.4`
  1. Migrate `DEPENDENCYMETRICS` data
  2. Drop old `DEPENDENCYMETRICS` table
  3. Rename table and index
* Changeset `v5.6.0-23.5`
  1. Create foreign key constraints
  2. Drop `MIGRATE_METRICS_TO_PARTITIONS` procedure

That way, there are more frequent "checkpoints" and the migration doesn't have to start over from zero in case it fails to complete in time.

nscuro (changes requested):

apiserver/src/main/java/org/dependencytrack/tasks/maintenance/MetricsMaintenanceTask.java
```diff
@@ -76,6 +77,9 @@ private record Statistics(
     private Statistics informLocked(final Handle jdbiHandle) {
         assertLocked();
 
+        // Create new partitions for today.
+        useJdbiHandle(handle -> handle.attach(MetricsDao.class).createMetricsPartitionsForToday());
```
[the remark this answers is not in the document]
sahibamittal: The task is scheduled to be run every day at 12AM. If it creates partitions for today and tomorrow, it would result in double attempts to create partitions for one day?

apiserver/src/main/resources/application.properties
```diff
@@ -1349,7 +1364,7 @@ task.component.metadata.maintenance.lock.min.duration=PT1M
 # @category: Task Scheduling
 # @type:     cron
 # @required
-task.metrics.maintenance.cron=0 */3 * * *
+task.metrics.maintenance.cron=0 12 * * *
```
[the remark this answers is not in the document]
sahibamittal: So it would be calling the script to create partitions multiple times (every hour)?

apiserver/src/main/java/org/dependencytrack/tasks/maintenance/MetricsMaintenanceTask.java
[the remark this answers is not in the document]
nscuro: > The task is scheduled to be run every day at 12AM.

It is not guaranteed to execute *exactly* at 12AM though. The task scheduler has an interval at which it checks for due tasks, which defaults to 1min:

https://github.com/DependencyTrack/hyades-apiserver/blob/a8aba4be3b8a768227448be735089a8dd277982b/apiserver/src/main/resources/application.properties#L915-L920

Which means that at worst, the task executes at 12:01AM.

Metrics maintenance is also executed in a single-threaded thread pool: 

https://github.com/DependencyTrack/hyades-apiserver/blob/a8aba4be3b8a768227448be735089a8dd277982b/apiserver/src/main/java/org/dependencytrack/event/EventSubsystemInitializer.java#L121-L124

So there *could* be other tasks queued that delay execution of metrics maintenance by a bit.

> it would result in double attempts to create partitions for one day?

Correct, but since you're using `CREATE TABLE IF NOT EXISTS`, attempts are cheap. If the respective tables already exist, nothing is executed.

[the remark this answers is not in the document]
sahibamittal: Right.. thanks for explaining 👍🏼

apiserver/src/main/resources/migration/changelog-v5.6.0.xml
sahibamittal: And for `DEPENDENCYMETRICS` it should be `PROJECT_ID, COMPONENT_ID, LAST_OCCURRENCE` or just `COMPONENT_ID, LAST_OCCURRENCE`

nscuro: Good question. We *do* query these by `PROJECT_ID` sometimes, for example when updating project metrics: https://github.com/DependencyTrack/hyades-apiserver/blob/dc9a1bf2e27374c6bb553c92fd281c0aac597a6e/apiserver/src/main/resources/migration/procedures/procedure_update-project-metrics.sql#L80-L83

So I think `PROJECT_ID, COMPONENT_ID, LAST_OCCURRENCE` would make sense.

apiserver/src/main/resources/application.properties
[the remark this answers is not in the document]
sahibamittal: I'm keeping it for every hour at 1st minute every day.

apiserver/src/main/resources/migration/changelog-v5.6.0.xml
sahibamittal: I'll have to keep the column "ID" though because if removed, migration is throwing exception `ERROR: INSERT has more expressions than target columns`

```
-- Insert data from existing table into this partition
                    EXECUTE format(
                        'INSERT INTO %I SELECT * FROM %I WHERE "LAST_OCCURRENCE" >= %L AND "LAST_OCCURRENCE" < %L;',
                        target_table,
                        source_table,
                        partition_date,
                        next_day
                    );
```

nscuro: I think this is "just" a matter of expanding `SELECT *` to all the columns we want (i.e. all except `ID`).

sahibamittal: I know I was avoiding the big query :D I'll edit it 👍🏼

nscuro: Two options I can think of:

* Define variables for each table with columns to select, then include them in the query via `%s`.
* Keep the `ID` column for the migration, but then immediately drop it afterwards. The downside is that dropping a column [doesn't really remove it](https://stackoverflow.com/a/15700185).

I would prefer the former. It's not pretty, but then again these procedures are temporary so they don't have to be.

WDYT? Any other options you see?

sahibamittal: I see online that the disk space for dropped column in Postgres can be reclaimed using [VACUUM](https://www.postgresql.org/docs/current/sql-vacuum.html) or [CLUSTER](https://www.postgresql.org/docs/current/sql-cluster.html) but it may slow it down if there is lots of migrated data in the table because whole table will get locked for it.

sahibamittal: To avoid any possible downtime, list of columns is the best way to do it. 👍🏼 Pushed the changes.

apiserver/src/test/java/org/dependencytrack/persistence/jdbi/ComponentDaoTest.java
```diff
@@ -159,6 +155,9 @@ public void testCascadeDeleteComponent() {
         assertThatNoException().isThrownBy(() -> qm.getObjectById(Vulnerability.class, vuln.getId()));
         assertThatNoException().isThrownBy(() -> qm.getObjectById(PolicyCondition.class, policyCondition.getId()));
         assertThatNoException().isThrownBy(() -> qm.getObjectById(Policy.class, policy.getId()));
+
+        // Ensure that metrics have been deleted.
+        assertThat(withJdbiHandle(handle ->  handle.attach(MetricsDao.class).getDependencyMetricsSince(component.getId(), Instant.now())).isEmpty());
```
nscuro, line 160: `Instant.now()` in this line may end up being a later point in time than `Instant.now()` when the `DependencyMetrics` record gets created. 

The `Instant` used here should be in the past to ensure it would technically include the metrics created earlier.

apiserver/src/test/java/org/dependencytrack/persistence/jdbi/MetricsDaoTest.java
```diff
@@ -56,25 +55,51 @@ public void after() {
...
+        component.setProject(project);
+        component.setName("Acme Component");
+        component.setVersion("1.0");
+        qm.createComponent(component, false);
+
+        var metrics = metricsDao.createDependencyMetrics(component.getId(), project.getId(), Instant.now(),
+                Instant.now(), 1, 0, 0, 0, 0, 0, 0);
+        assertThat(metrics).isNotNull();
+        assertThat(metrics.getCritical()).isEqualTo(1);
+    }
```
nscuro, line 88: We probably won't need tests for these since they're already only used in tests.

apiserver/src/test/java/org/dependencytrack/persistence/jdbi/ProjectDaoTest.java
```diff
@@ -229,6 +223,11 @@ public void testCascadeDeleteProject() {
         assertThat(notificationRule.getProjects()).isEmpty();
         qm.getPersistenceManager().refresh(policy);
         assertThat(policy.getProjects()).isEmpty();
+
+        // Ensure that metrics have been deleted.
+        MetricsDao dao = jdbiHandle.attach(MetricsDao.class);
+        assertThat(dao.getProjectMetricsSince(project.getId(), Instant.now())).isEmpty();
+        assertThat(dao.getDependencyMetricsSince(component.getId(), Instant.now())).isEmpty();
```
nscuro, line 230: The `Instant`s used here should be in the past, otherwise the method calls are guaranteed to never return anything.

apiserver/src/main/java/org/dependencytrack/persistence/DefaultObjectGenerator.java
```diff
@@ -362,4 +368,13 @@ private void loadDefaultNotificationPublishers(final QueryManager qm) {
...
         }
     }
+
+    /**
+     * Checks if metrics partitions exist for today and create if they don't exist.
+     */
+    void ensureMetricsPartitions() {
+        useJdbiHandle(handle -> handle.attach(MetricsDao.class).createMetricsPartitionsForDate(
+                LocalDate.now().toString(),
+                LocalDate.now().plusDays(1).toString()));
```
nscuro, line 378: Can we create partitions for *tomorrow* here as well, analogue to `MetricsMaintenanceTask`?

nscuro (changes requested):
Great work! 

I tested the migration in a setup with >5000 projects and ~4'000'000 components. Used the following script to populate metrics tables prior to executing the migrations:

<details>
<summary>Test Script</summary>

```sql
CREATE OR REPLACE FUNCTION generate_metrics_test_data(days_to_generate INTEGER)
RETURNS VOID AS $$
DECLARE
    base_date TIMESTAMP WITH TIME ZONE;
    batch_size INTEGER := 30; -- Process many days per batch to reduce overhead
    current_batch INTEGER := 0;
    batch_count INTEGER;
    start_day INTEGER;
    end_day INTEGER;
    project_count INTEGER;
    max_components_per_project INTEGER := 15; -- Maximum components per project
    sample_size INTEGER := 1000; -- Number of components to pre-select from the full set
    daily_variation FLOAT := 0.1; -- 10% day-to-day variation
BEGIN
    -- Get total project count for reporting
    SELECT COUNT(*) INTO project_count FROM public."PROJECT";
    RAISE NOTICE 'Processing % projects', project_count;
    
    -- Start from current date and go back the specified number of days
    base_date := NOW() - (days_to_generate || ' DAYS')::INTERVAL;
    
    -- Calculate number of batches
    batch_count := CEILING(days_to_generate::NUMERIC / batch_size);
    RAISE NOTICE 'Will process % days in % batches', days_to_generate, batch_count;
    
    -- Create a temporary table with a sampled subset of components
    CREATE TEMP TABLE sampled_components AS
    SELECT "ID" AS component_id
    FROM public."COMPONENT"
    ORDER BY RANDOM()
    LIMIT sample_size;
    
    -- Create index for faster joins
    CREATE INDEX idx_sampled_components ON sampled_components(component_id);
    
    -- Process in batches
    FOR current_batch IN 0..batch_count-1 LOOP
        start_day := current_batch * batch_size;
        end_day := LEAST((current_batch + 1) * batch_size - 1, days_to_generate - 1);
        
        RAISE NOTICE 'Processing days % to % (batch % of %)', 
            start_day + 1, end_day + 1, current_batch + 1, batch_count;
            
        -- FIRST PHASE: Generate core dependency metrics for this batch
        -- Create a temporary "blueprint" of project-component relationships
        -- with pre-calculated base metrics
        CREATE TEMP TABLE batch_dependency_blueprint AS
        WITH project_component_pairs AS (
            -- Assign a subset of components to each project using window functions
            SELECT 
                p."ID" AS project_id,
                c.component_id,
                -- Create a deterministic but seemingly random pattern per project
                ROW_NUMBER() OVER (PARTITION BY p."ID" ORDER BY RANDOM()) AS component_rank,
                -- Pre-compute base metrics that will be reused with small variations each day
                FLOOR(RANDOM() * RANDOM() * 3)::INTEGER AS base_critical,
                FLOOR(RANDOM() * RANDOM() * 5)::INTEGER AS base_high,
                FLOOR(RANDOM() * 8)::INTEGER AS base_medium,
                FLOOR(RANDOM() * 12)::INTEGER AS base_low,
                FLOOR(RANDOM() * 3)::INTEGER AS base_unassigned,
                FLOOR(RANDOM() * 5)::INTEGER AS base_policyviolations_license,
                FLOOR(RANDOM() * 4)::INTEGER AS base_policyviolations_security,
                FLOOR(RANDOM() * 3)::INTEGER AS base_policyviolations_operational
            FROM 
                public."PROJECT" p
                CROSS JOIN sampled_components c
        )
        SELECT 
            project_id,
            component_id,
            base_critical,
            base_high, 
            base_medium,
            base_low,
            base_unassigned,
            base_policyviolations_license,
            base_policyviolations_security,
            base_policyviolations_operational
        FROM 
            project_component_pairs
        WHERE 
            -- Each project gets 3-15 components (configurable via max_components_per_project)
            component_rank <= 3 + FLOOR(RANDOM() * (max_components_per_project - 3 + 1));
            
        -- Create index on the blueprint for faster access
        CREATE INDEX idx_batch_blueprint ON batch_dependency_blueprint(project_id, component_id);
            
        -- SECOND PHASE: Generate day variations and insert batch dependency metrics
        -- This approach uses days as the outer dimension to reduce memory pressure
        -- and improves performance by reducing the amount of data held in memory
        FOR day_offset IN start_day..end_day LOOP
            DECLARE
                metric_date TIMESTAMP WITH TIME ZONE := base_date + (day_offset || ' DAYS')::INTERVAL;
                day_fraction FLOAT := day_offset::FLOAT / GREATEST(1, days_to_generate)::FLOAT;
                -- Create a variation factor that slowly increases over time
                -- This creates more realistic pattern of changes over time
                trend_factor FLOAT := 1.0 + (day_fraction * 0.5); -- Max 50% increase by end
            BEGIN
                -- Insert dependency metrics for this specific day
                INSERT INTO public."DEPENDENCYMETRICS" (
                    "COMPONENT_ID", "PROJECT_ID", "CRITICAL", "HIGH", "MEDIUM", "LOW", 
                    "UNASSIGNED_SEVERITY", "VULNERABILITIES", "SUPPRESSED", 
                    "FINDINGS_TOTAL", "FINDINGS_AUDITED", "FINDINGS_UNAUDITED", 
                    "POLICYVIOLATIONS_TOTAL", "POLICYVIOLATIONS_FAIL", "POLICYVIOLATIONS_WARN", "POLICYVIOLATIONS_INFO", 
                    "POLICYVIOLATIONS_AUDITED", "POLICYVIOLATIONS_UNAUDITED", 
                    "POLICYVIOLATIONS_SECURITY_TOTAL", "POLICYVIOLATIONS_SECURITY_AUDITED", "POLICYVIOLATIONS_SECURITY_UNAUDITED", 
                    "POLICYVIOLATIONS_LICENSE_TOTAL", "POLICYVIOLATIONS_LICENSE_AUDITED", "POLICYVIOLATIONS_LICENSE_UNAUDITED", 
                    "POLICYVIOLATIONS_OPERATIONAL_TOTAL", "POLICYVIOLATIONS_OPERATIONAL_AUDITED", "POLICYVIOLATIONS_OPERATIONAL_UNAUDITED", 
                    "RISKSCORE", "FIRST_OCCURRENCE", "LAST_OCCURRENCE"
                )
                WITH daily_variations AS (
                    -- Apply daily variations to base metrics
                    SELECT
                        component_id,
                        project_id,
                        -- Apply variation and trend to each metric
                        -- Use GREATEST(1, ...) to ensure reasonable values
                        GREATEST(0, ROUND(base_critical * trend_factor * 
                            (1.0 + CASE WHEN RANDOM() > 0.5 THEN daily_variation ELSE -daily_variation END))::INTEGER) AS critical,
                        GREATEST(0, ROUND(base_high * trend_factor * 
                            (1.0 + CASE WHEN RANDOM() > 0.5 THEN daily_variation ELSE -daily_variation END))::INTEGER) AS high,
                        GREATEST(0, ROUND(base_medium * trend_factor * 
                            (1.0 + CASE WHEN RANDOM() > 0.5 THEN daily_variation ELSE -daily_variation END))::INTEGER) AS medium,
                        GREATEST(0, ROUND(base_low * trend_factor * 
                            (1.0 + CASE WHEN RANDOM() > 0.5 THEN daily_variation ELSE -daily_variation END))::INTEGER) AS low,
                        GREATEST(0, ROUND(base_unassigned * trend_factor * 
                            (1.0 + CASE WHEN RANDOM() > 0.5 THEN daily_variation ELSE -daily_variation END))::INTEGER) AS unassigned_severity,
                        GREATEST(0, ROUND(base_policyviolations_license * trend_factor * 
                            (1.0 + CASE WHEN RANDOM() > 0.5 THEN daily_variation ELSE -daily_variation END))::INTEGER) AS policyviolations_license_total,
                        GREATEST(0, ROUND(base_policyviolations_security * trend_factor * 
                            (1.0 + CASE WHEN RANDOM() > 0.5 THEN daily_variation ELSE -daily_variation END))::INTEGER) AS policyviolations_security_total,
                        GREATEST(0, ROUND(base_policyviolations_operational * trend_factor * 
                            (1.0 + CASE WHEN RANDOM() > 0.5 THEN daily_variation ELSE -daily_variation END))::INTEGER) AS policyviolations_operational_total
                    FROM 
                        batch_dependency_blueprint
                ),
                suppressed_values AS (
                    -- Calculate suppressed values
                    SELECT
                        component_id,
                        project_id,
                        critical,
                        high,
                        medium,
                        low,
                        unassigned_severity,
                        policyviolations_license_total,
                        policyviolations_security_total,
                        policyviolations_operational_total,
                        -- Calculate suppressed as percentage of total vulnerabilities
                        GREATEST(0, ROUND((critical + high + medium + low + unassigned_severity) * 0.2 * RANDOM())::INTEGER) AS suppressed
                    FROM
                        daily_variations
                ),
                audited_values AS (
                    -- Calculate audited values for policy violations
                    SELECT
                        component_id,
                        project_id,
                        critical,
                        high,
                        medium,
                        low,
                        unassigned_severity,
                        suppressed,
                        policyviolations_license_total,
                        GREATEST(0, ROUND(policyviolations_license_total * RANDOM())::INTEGER) AS policyviolations_license_audited,
                        policyviolations_security_total,
                        GREATEST(0, ROUND(policyviolations_security_total * RANDOM())::INTEGER) AS policyviolations_security_audited,
                        policyviolations_operational_total,
                        GREATEST(0, ROUND(policyviolations_operational_total * RANDOM())::INTEGER) AS policyviolations_operational_audited
                    FROM
                        suppressed_values
                ),
                final_metrics AS (
                    -- Calculate all derived metrics
                    SELECT
                        component_id,
                        project_id,
                        critical,
                        high,
                        medium,
                        low,
                        unassigned_severity,
                        suppressed,
                        -- Total vulnerabilities
                        (critical + high + medium + low + unassigned_severity - suppressed) AS vulnerabilities,
                        
                        -- Policy violation totals
                        policyviolations_license_total,
                        policyviolations_license_audited,
                        policyviolations_license_total - policyviolations_license_audited AS policyviolations_license_unaudited,
                        
                        policyviolations_security_total,
                        policyviolations_security_audited,
                        policyviolations_security_total - policyviolations_security_audited AS policyviolations_security_unaudited,
                        
                        policyviolations_operational_total,
                        policyviolations_operational_audited,
                        policyviolations_operational_total - policyviolations_operational_audited AS policyviolations_operational_unaudited,
                        
                        -- Combined policy violations
                        (policyviolations_license_total + policyviolations_security_total + policyviolations_operational_total) AS policyviolations_total,
                        
                        -- Policy violations by severity (using percentages of total)
                        GREATEST(0, ROUND((policyviolations_license_total + policyviolations_security_total + policyviolations_operational_total) * 0.3 * RANDOM())::INTEGER) AS policyviolations_fail,
                        
                        -- Total findings (vulnerabilities + policy violations)
                        (critical + high + medium + low + unassigned_severity - suppressed + 
                        policyviolations_license_total + policyviolations_security_total + policyviolations_operational_total) AS findings_total,
                        
                        -- Risk score based on weighted severity
                        (critical * 10.0 + high * 5.0 + medium * 2.0 + low * 1.0) * 
                            (1.0 - (suppressed::FLOAT / GREATEST(1, critical + high + medium + low + unassigned_severity))) AS riskscore
                    FROM
                        audited_values
                )
                SELECT
                    component_id,
                    project_id,
                    critical,
                    high,
                    medium,
                    low,
                    unassigned_severity,
                    vulnerabilities,
                    suppressed,
                    findings_total,
                    -- Findings audited (60% random)
                    GREATEST(0, ROUND(findings_total * 0.6 * RANDOM())::INTEGER) AS findings_audited,
                    -- Findings unaudited (remainder)
                    findings_total - GREATEST(0, ROUND(findings_total * 0.6 * RANDOM())::INTEGER) AS findings_unaudited,
                    
                    policyviolations_total,
                    policyviolations_fail,
                    -- Policy violations by severity (warn & info)
                    GREATEST(0, ROUND((policyviolations_total - policyviolations_fail) * 0.5 * RANDOM())::INTEGER) AS policyviolations_warn,
                    policyviolations_total - policyviolations_fail - 
                        GREATEST(0, ROUND((policyviolations_total - policyviolations_fail) * 0.5 * RANDOM())::INTEGER) AS policyviolations_info,
                    
                    -- Policy violations audited
                    (policyviolations_license_audited + policyviolations_security_audited + policyviolations_operational_audited) AS policyviolations_audited,
                    -- Policy violations unaudited
                    policyviolations_total - (policyviolations_license_audited + policyviolations_security_audited + policyviolations_operational_audited) AS policyviolations_unaudited,
                    
                    -- Security policy violations
                    policyviolations_security_total,
                    policyviolations_security_audited,
                    policyviolations_security_unaudited,
                    
                    -- License policy violations
                    policyviolations_license_total,
                    policyviolations_license_audited,
                    policyviolations_license_unaudited,
                    
                    -- Operational policy violations
                    policyviolations_operational_total,
                    policyviolations_operational_audited,
                    policyviolations_operational_unaudited,
                    
                    riskscore,
                    metric_date AS first_occurrence,
                    metric_date AS last_occurrence
                FROM
                    final_metrics;
                
                -- Calculate and insert project metrics for this day in one operation
                INSERT INTO public."PROJECTMETRICS" (
                    "PROJECT_ID", "COMPONENTS", "VULNERABLECOMPONENTS", "CRITICAL", "HIGH", "MEDIUM", "LOW", 
                    "UNASSIGNED_SEVERITY", "VULNERABILITIES", "SUPPRESSED", 
                    "FINDINGS_TOTAL", "FINDINGS_AUDITED", "FINDINGS_UNAUDITED", 
                    "POLICYVIOLATIONS_TOTAL", "POLICYVIOLATIONS_FAIL", "POLICYVIOLATIONS_WARN", "POLICYVIOLATIONS_INFO", 
                    "POLICYVIOLATIONS_AUDITED", "POLICYVIOLATIONS_UNAUDITED", 
                    "POLICYVIOLATIONS_SECURITY_TOTAL", "POLICYVIOLATIONS_SECURITY_AUDITED", "POLICYVIOLATIONS_SECURITY_UNAUDITED", 
                    "POLICYVIOLATIONS_LICENSE_TOTAL", "POLICYVIOLATIONS_LICENSE_AUDITED", "POLICYVIOLATIONS_LICENSE_UNAUDITED", 
                    "POLICYVIOLATIONS_OPERATIONAL_TOTAL", "POLICYVIOLATIONS_OPERATIONAL_AUDITED", "POLICYVIOLATIONS_OPERATIONAL_UNAUDITED", 
                    "RISKSCORE", "FIRST_OCCURRENCE", "LAST_OCCURRENCE"
                )
                SELECT
                    "PROJECT_ID" AS project_id,
                    COUNT(*) AS components,
                    COUNT(CASE WHEN "VULNERABILITIES" > 0 THEN 1 END) AS vulnerablecomponents,
                    SUM("CRITICAL") AS critical,
                    SUM("HIGH") AS high,
                    SUM("MEDIUM") AS medium,
                    SUM("LOW") AS low,
                    SUM("UNASSIGNED_SEVERITY") AS unassigned_severity,
                    SUM("VULNERABILITIES") AS vulnerabilities,
                    SUM("SUPPRESSED") AS suppressed,
                    SUM("FINDINGS_TOTAL") AS findings_total,
                    SUM("FINDINGS_AUDITED") AS findings_audited,
                    SUM("FINDINGS_UNAUDITED") AS findings_unaudited,
                    SUM("POLICYVIOLATIONS_TOTAL") AS policyviolations_total,
                    SUM("POLICYVIOLATIONS_FAIL") AS policyviolations_fail,
                    SUM("POLICYVIOLATIONS_WARN") AS policyviolations_warn,
                    SUM("POLICYVIOLATIONS_INFO") AS policyviolations_info,
                    SUM("POLICYVIOLATIONS_AUDITED") AS policyviolations_audited,
                    SUM("POLICYVIOLATIONS_UNAUDITED") AS policyviolations_unaudited,
                    SUM("POLICYVIOLATIONS_SECURITY_TOTAL") AS policyviolations_security_total,
                    SUM("POLICYVIOLATIONS_SECURITY_AUDITED") AS policyviolations_security_audited,
                    SUM("POLICYVIOLATIONS_SECURITY_UNAUDITED") AS policyviolations_security_unaudited,
                    SUM("POLICYVIOLATIONS_LICENSE_TOTAL") AS policyviolations_license_total,
                    SUM("POLICYVIOLATIONS_LICENSE_AUDITED") AS policyviolations_license_audited,
                    SUM("POLICYVIOLATIONS_LICENSE_UNAUDITED") AS policyviolations_license_unaudited,
                    SUM("POLICYVIOLATIONS_OPERATIONAL_TOTAL") AS policyviolations_operational_total,
                    SUM("POLICYVIOLATIONS_OPERATIONAL_AUDITED") AS policyviolations_operational_audited,
                    SUM("POLICYVIOLATIONS_OPERATIONAL_UNAUDITED") AS policyviolations_operational_unaudited,
                    SUM("RISKSCORE") AS riskscore,
                    metric_date AS first_occurrence,
                    metric_date AS last_occurrence
                FROM
                    public."DEPENDENCYMETRICS"
                WHERE
                    "LAST_OCCURRENCE" = metric_date
                GROUP BY
                    "PROJECT_ID";
                
                -- Calculate and insert portfolio metrics for this day in one operation
                INSERT INTO public."PORTFOLIOMETRICS" (
                    "PROJECTS", "VULNERABLEPROJECTS", "COMPONENTS", "VULNERABLECOMPONENTS", "CRITICAL", "HIGH", "MEDIUM", "LOW", 
                    "UNASSIGNED_SEVERITY", "VULNERABILITIES", "SUPPRESSED", 
                    "FINDINGS_TOTAL", "FINDINGS_AUDITED", "FINDINGS_UNAUDITED", 
                    "POLICYVIOLATIONS_TOTAL", "POLICYVIOLATIONS_FAIL", "POLICYVIOLATIONS_WARN", "POLICYVIOLATIONS_INFO", 
                    "POLICYVIOLATIONS_AUDITED", "POLICYVIOLATIONS_UNAUDITED", 
                    "POLICYVIOLATIONS_SECURITY_TOTAL", "POLICYVIOLATIONS_SECURITY_AUDITED", "POLICYVIOLATIONS_SECURITY_UNAUDITED", 
                    "POLICYVIOLATIONS_LICENSE_TOTAL", "POLICYVIOLATIONS_LICENSE_AUDITED", "POLICYVIOLATIONS_LICENSE_UNAUDITED", 
                    "POLICYVIOLATIONS_OPERATIONAL_TOTAL", "POLICYVIOLATIONS_OPERATIONAL_AUDITED", "POLICYVIOLATIONS_OPERATIONAL_UNAUDITED", 
                    "RISKSCORE", "FIRST_OCCURRENCE", "LAST_OCCURRENCE"
                )
                SELECT
                    COUNT(*) AS projects,
                    COUNT(CASE WHEN "VULNERABILITIES" > 0 THEN 1 END) AS vulnerableprojects,
                    SUM("COMPONENTS") AS components,
                    SUM("VULNERABLECOMPONENTS") AS vulnerablecomponents,
                    SUM("CRITICAL") AS critical,
                    SUM("HIGH") AS high,
                    SUM("MEDIUM") AS medium,
                    SUM("LOW") AS low,
                    SUM("UNASSIGNED_SEVERITY") AS unassigned_severity,
                    SUM("VULNERABILITIES") AS vulnerabilities,
                    SUM("SUPPRESSED") AS suppressed,
                    SUM("FINDINGS_TOTAL") AS findings_total,
                    SUM("FINDINGS_AUDITED") AS findings_audited,
                    SUM("FINDINGS_UNAUDITED") AS findings_unaudited,
                    SUM("POLICYVIOLATIONS_TOTAL") AS policyviolations_total,
                    SUM("POLICYVIOLATIONS_FAIL") AS policyviolations_fail,
                    SUM("POLICYVIOLATIONS_WARN") AS policyviolations_warn,
                    SUM("POLICYVIOLATIONS_INFO") AS policyviolations_info,
                    SUM("POLICYVIOLATIONS_AUDITED") AS policyviolations_audited,
                    SUM("POLICYVIOLATIONS_UNAUDITED") AS policyviolations_unaudited,
                    SUM("POLICYVIOLATIONS_SECURITY_TOTAL") AS policyviolations_security_total,
                    SUM("POLICYVIOLATIONS_SECURITY_AUDITED") AS policyviolations_security_audited,
                    SUM("POLICYVIOLATIONS_SECURITY_UNAUDITED") AS policyviolations_security_unaudited,
                    SUM("POLICYVIOLATIONS_LICENSE_TOTAL") AS policyviolations_license_total,
                    SUM("POLICYVIOLATIONS_LICENSE_AUDITED") AS policyviolations_license_audited,
                    SUM("POLICYVIOLATIONS_LICENSE_UNAUDITED") AS policyviolations_license_unaudited,
                    SUM("POLICYVIOLATIONS_OPERATIONAL_TOTAL") AS policyviolations_operational_total,
                    SUM("POLICYVIOLATIONS_OPERATIONAL_AUDITED") AS policyviolations_operational_audited,
                    SUM("POLICYVIOLATIONS_OPERATIONAL_UNAUDITED") AS policyviolations_operational_unaudited,
                    SUM("RISKSCORE") AS riskscore,
                    metric_date AS first_occurrence,
                    metric_date AS last_occurrence
                FROM
                    public."PROJECTMETRICS"
                WHERE
                    "LAST_OCCURRENCE" = metric_date;
                    
                -- Progress indicator within the batch
                IF day_offset % 10 = 0 THEN
                    RAISE NOTICE 'Completed day % of %', day_offset + 1, days_to_generate;
                END IF;
            END;
        END LOOP;
        
        -- Clean up the blueprint table after processing the batch
        DROP TABLE IF EXISTS batch_dependency_blueprint;
    END LOOP;
    
    -- Clean up
    DROP TABLE IF EXISTS sampled_components;
    
    RAISE NOTICE 'Successfully generated metrics data for % days', days_to_generate;
END;
$$ LANGUAGE plpgsql;

SELECT generate_metrics_test_data(180);
```

</details>

Migration went smooth and fairly fast as well. I then changed the metrics retention from 90 to 30 days and waited for the maintenance task to run. It dropped 60 partitions per table in just barely over 1sec.

There are only some minor code changes I'd like to see, but this is really good already!

sahibamittal:
> Great work!
> 
> I tested the migration in a setup with >5000 projects and ~4'000'000 components. Used the following script to populate metrics tables prior to executing the migrations:
> 
> Test Script
> Migration went smooth and fairly fast as well. I then changed the metrics retention from 90 to 30 days and waited for the maintenance task to run. It dropped 60 partitions per table in just barely over 1sec.
> 
> There are only some minor code changes I'd like to see, but this is really good already!

Thankyou for the thorough load testing! 💯

apiserver/src/main/java/org/dependencytrack/persistence/jdbi/MetricsDao.java
```diff
@@ -87,4 +103,144 @@ public interface MetricsDao {
...
+                        partition_name,
+                        today,
+                        tomorrow
+                    );
+                END IF;
+            END LOOP;
+        END;
+        $$;
+    """)
+    void createMetricsPartitionsForDate(@Define("targetDate") String targetDate, @Define("nextDate") String nextDate);
```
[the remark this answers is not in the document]
sahibamittal: It seems the `LocalDate` parameter cannot be injected into the `DO` block.. like 
`today DATE := cast(:targetDate as DATE)` OR `today DATE := :targetDate` 
with `@Bind("targetDate") LocalDate targetDate`.

apiserver/src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
```diff
@@ -826,15 +806,15 @@ public void getProjectsConciseWithLatestMetricsTest() {
...
+                      "policyViolationsFail": 0,
+                      "policyViolationsInfo": 0,
+                      "policyViolationsLicenseTotal": 0,
+                      "policyViolationsOperationalTotal": 0,
+                      "policyViolationsSecurityTotal": 0,
+                      "policyViolationsTotal": 0,
+                      "policyViolationsWarn": 0,
                       "inheritedRiskScore": 13.13,
-                      "unassigned": 14,
+                      "unassigned": 0,
```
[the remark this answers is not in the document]
sahibamittal: When I moved the logic of persistence to JDBI `createPortfolioMetrics` I kept only `not_null` fields in the method to keep it short since it's only being used in couple tests.

apiserver/src/main/java/org/dependencytrack/model/PortfolioMetrics.java
```diff
@@ -45,11 +43,6 @@ public class PortfolioMetrics implements Serializable {
 
     private static final long serialVersionUID = -7690624184866776922L;
 
-    @PrimaryKey
-    @Persistent(valueStrategy = IdGeneratorStrategy.NATIVE)
```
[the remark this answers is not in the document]
sahibamittal: Same for Project and Dependency metrics?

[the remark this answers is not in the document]
nscuro: Yes please.

apiserver/src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
```diff
@@ -1309,15 +1269,15 @@ public void getProjectChildrenConciseWithLatestMetricsTest() {
...
```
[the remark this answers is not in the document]
sahibamittal: When I moved the logic of persistence to JDBI `createProjectMetrics` I kept only `not_null` fields in the method to keep it short since it's only being used in couple tests. Do you want me add these fields in Insert query?

apiserver/src/test/java/org/dependencytrack/metrics/MetricsUtil.java
```diff
@@ -0,0 +1,89 @@
...
+import org.dependencytrack.model.ProjectMetrics;
+import org.jdbi.v3.sqlobject.SqlObject;
+import org.jdbi.v3.sqlobject.config.RegisterBeanMapper;
+import org.jdbi.v3.sqlobject.customizer.BindBean;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public interface MetricsUtil extends SqlObject {
```
nscuro, line 32: Nit: Having a DAO named `Util` is a bit odd, perhaps name it `MetricsTestDao` instead and move it to `persistence.jdbi` instead?

apiserver/src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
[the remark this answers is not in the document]
nscuro: No current way is fine, was just wondering where the numbers went :D

apiserver/src/test/java/org/dependencytrack/resources/v1/ProjectResourceTest.java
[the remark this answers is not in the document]
nscuro: See other comment. It's OK like this, thanks for explaining.

nscuro (commented):
Few last tweaks, but otherwise good to go!

nscuro (approved):
Great work again, @sahibamittal! Please remember to update `schema.sql` in the `hyades` repo accordingly!

PR #1251, opened 2025-06-05
Enforce permissions at project level

jhoward-lm:
### Description

This PR introduces project-scoped permissions by leveraging the `USER_PROJECT_EFFECTIVE_PERMISSIONS` table and the concept of roles.

### Addressed Issue

resolves DependencyTrack/hyades#1075
resolves DependencyTrack/hyades#1422

### Checklist

- [ ] ~This PR fixes a defect, and I have provided tests to verify that the fix is effective~


jhoward-lm:
@nscuro let me know if you have a preferred alternative to the `AuthorizationFilter` and `AuthorizationFeature` that were added under `apiserver`

apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
```diff
@@ -134,7 +135,7 @@ public class BomResource extends AbstractApiResource {
                     content = @Content(schema = @Schema(implementation = ProblemDetails.class), mediaType = ProblemDetails.MEDIA_TYPE_JSON)),
             @ApiResponse(responseCode = "404", description = "The project could not be found")
     })
-    @PermissionRequired(Permissions.Constants.VIEW_PORTFOLIO)
+    @DisableAuthorization
```
nscuro, line 138: Unfortunately for the mid-term we'll need to continue supporting `@PermissionRequired`. This is relevant for deployments that do not use portfolio ACL, where removing `@PermissionRequired` effectively removes AuthZ entirely.

What this means is that Alpine's `AuthorizationFilter` will need to be aware of the new annotation, and omit AuthZ checks when it's present. If the filter needs to access API server code, consider moving the filter to the `apiserver` module.

alpine/alpine-server/src/main/java/alpine/server/auth/DisableAuthorization.java
```diff
@@ -0,0 +1,40 @@
...
+ * disable authorization checks when attempting to access the method.
+ *
+ * @author Jonathan Howard
+ * @since 5.6.0
+ */
+@Inherited
+@Retention(value = RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+@Documented
+public @interface DisableAuthorization {
```
nscuro, line 39: Would be good to find a better name for this before the PR gets merged. Something that expresses that AuthZ is not disabled entirely, but just performed in a different place (if portfolio ACL is enabled).

Alternatively, something like the `@AuthorizeProject` annotation we chatted about, that makes the "granular AuthZ" more obvious, could help.

jhoward-lm: I'm fine with renaming it; how would you feel about something like `@ProjectAccessRequired` or `@ProjectAuthorizationRequired` for consistency with `@PermissionRequired`?

jhoward-lm: Renamed to `ProjectAccessRequired`

nscuro (approved):
Tested and works as expected.

A caveat being that this so far only covers *read* access. More work will be needed to make this work for permissions outside of `VIEW_PORTFOLIO` (the PR description currently does not reflect this). As such this doesn't yet resolve https://github.com/DependencyTrack/hyades/issues/1075.

I think in the next step we need to modify `has_user_project_access` so it takes the permission(s) to check for as an argument.

I am happy to merge this PR and have the remaining work be done in a separate one. Up to you.

jhoward-lm:
> Tested and works as expected.

Thank you for testing!

> A caveat being that this so far only covers _read_ access. More work will be needed to make this work for permissions outside of `VIEW_PORTFOLIO` (the PR description currently does not reflect this). As such this doesn't yet resolve [DependencyTrack/hyades#1075](https://github.com/DependencyTrack/hyades/issues/1075).
> 
> I think in the next step we need to modify `has_user_project_access` so it takes the permission(s) to check for as an argument.
> 
> I am happy to merge this PR and have the remaining work be done in a separate one. Up to you.

I think a separate PR makes sense, so let's go ahead and merge this one. Thanks again!

PR #1283, opened 2025-06-19
Add pagination to `/api/v1/team` -> `getTeams`

sahibamittal:
### Description

Add pagination to `/api/v1/team` -> `getTeams`

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/1798

### Additional Details

Frontend changes not required. Pagination is already `true` for `teams`.

### Checklist

### DIFF
--- a/alpine/alpine-infra/src/main/java/alpine/persistence/AlpineQueryManager.java
+++ b/alpine/alpine-infra/src/main/java/alpine/persistence/AlpineQueryManager.java
@@ -581,14 +581,11 @@ public class AlpineQueryManager extends AbstractAlpineQueryManager {
      * @return a List of Teams
      * @since 1.0.0
      */
-    public List<Team> getTeams() {
+    public PaginatedResult getTeams() {
         final Query<Team> query = pm.newQuery(Team.class);
         query.getFetchPlan().addGroup(Team.FetchGroup.ALL.name());
         query.setOrdering("name asc");
-        if (pagination != null && pagination.isPaginated()) {
-            query.setRange(pagination.getOffset(), pagination.getOffset() + pagination.getLimit());
-        }
-        return executeAndCloseList(query);
+        return execute(query);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


alpine/alpine-infra/src/main/java/alpine/persistence/AlpineQueryManager.java
```diff
@@ -586,6 +585,9 @@ public List<Team> getTeams() {
         final Query<Team> query = pm.newQuery(Team.class);
```
nscuro, line 585: The method will need to return `PaginatedResult`, otherwise the API cannot provide the `X-Total-Count` header. You can simply use this `execute` method to cater to all this:

https://github.com/DependencyTrack/hyades-apiserver/blob/ad45a0ef8e3c2041945fc8dd1e1a489a8cc0f4fb/alpine/alpine-infra/src/main/java/alpine/persistence/AbstractAlpineQueryManager.java#L148-L162

I also think the frontend has a search box in the teams view. In that case we need to make sure that `filter` is being used (which is populated from the request's `searchText` query parameter). Something like this:

https://github.com/DependencyTrack/hyades-apiserver/blob/ad45a0ef8e3c2041945fc8dd1e1a489a8cc0f4fb/apiserver/src/main/java/org/dependencytrack/persistence/VulnerabilityQueryManager.java#L312-L316

nscuro: Does it find teams on a page that is isn't the current one? Say you have 26 teams named `A` to `Z`, you sort them by name and set the page size to `10`. Does it yield any results when you search for `Z`?

sahibamittal: Yes, like in SS below, `Team 8` is listed on second page and when I filter by `8` on first page, it appears.

Screenshot 2025-06-23 at 13 38 10
Screenshot 2025-06-23 at 13 38 21
Screenshot 2025-06-23 at 13 38 34

sahibamittal: It seems it's been done dynamically below : 
```
public PaginatedResult execute(final Query<?> query, final Object... parameters) {
        final long count = getCount(query, parameters);
        decorate(query);
        return new PaginatedResult()
                .objects(executeAndCloseWithArray(query, parameters))
                .total(count);
    }
```

PR #1052, opened 2025-02-04
Add OIDC authentication customizer for GitLab

jhoward-lm:
### Description

This PR adds an implementation of the `OidcAuthenticationCustomizer` service provider interface specific to GitLab. It is meant to synchronize a user's projects and roles/max access levels per project within GitLab to a Hyades instance.

Depends on stevespringett/Alpine#755 and #1069.

> [!NOTE]
> WIP: a rough to-do list with outstanding items is in a comment block in `GitLabSyncTask.java`

### Addressed Issue

Closes DependencyTrack/hyades#1632

### Additional Details

This PR is a request for early review while still in development in order to change direction if necessary.

### Checklist

- [X] I have read and understand the [contributing guidelines]
- [ ] ~This PR fixes a defect, and I have provided tests to verify that the fix is effective~
- [ ] This PR implements an enhancement, and I have provided tests to verify that it works as intended
- [ ] This PR introduces changes to the database model, and I have updated the [migration changelog] accordingly
- [ ] This PR introduces new or alters existing behavior, and I have updated the [documentation] accordingly

[contributing guidelines]: ../CONTRIBUTING.md#pull-requests
[documentation]: https://dependencytrack.github.io/hyades/latest/development/documentation/
[migration changelog]: https://dependencytrack.github.io/hyades/latest/development/database-migrations/

PR #1325, opened 2025-07-16
Gitlab integration

jhoward-lm:
### Description

This PR adds an integration for authentication and authorization using GitLab as an OIDC issuer. Its purpose is to synchronize a user's projects and roles/max access levels per project within GitLab to a DependencyTrack instance.

It includes:

- An implementation of the `OidcAuthenticationCustomizer` service provider interface specific to GitLab
- `GitLabSyncer` integration class
- `GitLabClient` class for querying the GraphQL API to retrieve user's projects and access levels per project
- Updates to resources to support integration functionality, in particular the ability to push an SBOM to a DependencyTrack instance from a GitLab CI/CD pipeline using the CI job's ID token

Supersedes #1052

### Checklist

- [ ] ~This PR fixes a defect, and I have provided tests to verify that the fix is effective~
- [X] This PR implements an enhancement, and I have provided tests to verify that it works as intended
- [ ] ~This PR introduces new or alters existing behavior, and I have updated the [documentation] accordingly~


apiserver/src/main/java/org/dependencytrack/persistence/jdbi/mapping/ProjectRowMapper.java
nscuro, line 1: This class seems to be unused.

apiserver/src/main/java/org/dependencytrack/integrations/gitlab/GitLabRole.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.dependencytrack.auth.Permissions;
+
+/**
+ * Definitions of access levels/roles as defined by GitLab.
+ */
+public enum GitLabRole {
```
nscuro, line 31: Can we add a link or reference here where GitLab defines these roles?

apiserver/src/main/java/org/dependencytrack/integrations/gitlab/GitLabRole.java
```diff
@@ -0,0 +1,103 @@
...
+            Permissions.Constants.ACCESS_MANAGEMENT_READ,
+            Permissions.Constants.ACCESS_MANAGEMENT_UPDATE,
+            Permissions.Constants.ACCESS_MANAGEMENT_DELETE,
+            Permissions.Constants.SYSTEM_CONFIGURATION,
+            Permissions.Constants.SYSTEM_CONFIGURATION_CREATE,
+            Permissions.Constants.SYSTEM_CONFIGURATION_READ,
+            Permissions.Constants.SYSTEM_CONFIGURATION_UPDATE,
+            Permissions.Constants.SYSTEM_CONFIGURATION_DELETE,
+            Permissions.Constants.TAG_MANAGEMENT,
+            Permissions.Constants.TAG_MANAGEMENT_DELETE));
```
nscuro, line 71: These permissions are currently not scoped to projects, but apply to the entire system. This would make all project owners also system administrators, which is not what we want.

apiserver/src/main/java/org/dependencytrack/integrations/gitlab/GitLabSyncer.java
```diff
@@ -0,0 +1,128 @@
...
+    @Override
+    public String description() {
+        return "Synchronizes user permissions from connected GitLab instance";
+    }
+
+    @Override
+    public boolean isEnabled() {
+        final ConfigProperty enabled = qm.getConfigProperty(INTEGRATIONS_GROUP, GITLAB_ENABLED.getPropertyName());
+
+        return enabled != null && Boolean.parseBoolean(enabled.getPropertyValue());
```
nscuro, line 78:
```suggestion
        return qm.isEnabled(GITLAB_ENABLED);
```

apiserver/src/main/java/org/dependencytrack/integrations/gitlab/GitLabSyncer.java
```diff
@@ -0,0 +1,128 @@
...
+
+    private List<Project> createProjects(List<GitLabProject> gitLabProjects) {
+        List<Project> projects = new ArrayList<>();
+
+        Map<GitLabRole, Role> roleMap = Arrays.stream(GitLabRole.values())
+                .collect(Collectors.toMap(Function.identity(), role -> qm.getRoleByName(role.getDescription())));
+
+        for (var gitLabProject : gitLabProjects) {
+            final Lock lock = locks.get(gitLabProject.getFullPath());
+            lock.lock();
```
nscuro, line 101: Let's switch this to use advisory locks instead. That way the locks are effective across API server instances:

https://github.com/DependencyTrack/hyades-apiserver/blob/db00ad60729c0d88e34ab856a62a0459d2aa648b/apiserver/src/main/java/org/dependencytrack/persistence/QueryManager.java#L1751-L1767

The implementation above is non-blocking and doesn't wait for the lock to become available. If that behavior is desired we'll need another implementation that calls `pg_advisory_xact_lock` instead.

apiserver/src/main/java/org/dependencytrack/resources/v1/PermissionResource.java
```diff
@@ -165,25 +170,37 @@ public Response removePermissionFromUser(
...
         try (QueryManager qm = new QueryManager()) {
             return qm.callInTransaction(() -> {
-                User principal = qm.getUser(username);
-                if (principal == null) {
+                User user = qm.getUser(username, (Class<? extends User>) switch (StringUtils.defaultString(type).toLowerCase()) {
+                    case "managed" -> ManagedUser.class;
+                    case "ldap" -> LdapUser.class;
+                    case "oidc" -> OidcUser.class;
+                    default -> User.class;
+                });
```
nscuro, line 182: The concrete user type seems to be irrelevant for what this method is doing.

Does this pre-date [ADR-006](https://github.com/DependencyTrack/hyades/blob/2f423efb8478454a54866f15e60672bc3975b251/docs/architecture/decisions/006-consolidate-user-tables.md) per chance, and thus had to deal with the same username existing for multiple user types? If so, I think this can be removed since uniqueness of usernames is now enforced across all types.

apiserver/src/main/java/org/dependencytrack/resources/v1/PermissionResource.java
```diff
@@ -378,29 +398,25 @@ public Response removePermissionFromTeam(
...
+            @Parameter(description = "A username and valid list permission") @Valid final UserPermissionsSetRequest request) {
         try (QueryManager qm = new QueryManager()) {
             return qm.callInTransaction(() -> {
-                User user = qm.getUser(request.username());
+                User user = qm.getUser(request.username(), (Class<? extends User>) switch (StringUtils.defaultString(request.userType()).toLowerCase()) {
+                    case "managed" -> ManagedUser.class;
+                    case "ldap" -> LdapUser.class;
+                    case "oidc" -> OidcUser.class;
+                    default -> User.class;
+                });
```
nscuro, line 409: Same as above, I assume this can be reverted.

apiserver/src/main/java/org/dependencytrack/resources/v1/PermissionResource.java
```diff
@@ -471,4 +477,47 @@ public Response setTeamPermissions(@Parameter(description = "Team UUID and reque
...
+            if (role.getPermissions().equals(requestedPermissions))
+                return Response.notModified().entity("Role already has selected permission(s).").build();
+
+            role.setPermissions(requestedPermissions);
+            role = qm.persist(role);
+
+            super.logSecurityEvent(LOGGER, SecurityMarkers.SECURITY_AUDIT,
+                    "Set permissions for role: %s / permissions: %s"
+                            .formatted(role.getName(), permissionNames));
+            return Response.ok(role).build();
```
nscuro, line 519: Please wrap this in a transaction using `qm.callInTransaction`.

apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
```diff
@@ -360,6 +368,120 @@ public Response uploadBom(@Parameter(required = true) BomSubmitRequest request)
...
+    })
+    @PermissionRequired(Permissions.Constants.BOM_UPLOAD)
+    @ResourceAccessRequired
+    public Response uploadBomGitLab(
+            @FormDataParam("gitLabToken") String idToken,
+            @FormDataParam("bom") String bom,
+            @FormDataParam("isLatest") @DefaultValue("false") boolean isLatest) {
+
+        try (QueryManager qm = new QueryManager()) {
+            Function<ConfigPropertyConstants, ConfigProperty> propertyGetter = cpc -> qm.getConfigProperty(
```
nscuro, line 390: Please wrap this in a transaction using `qm.callInTransaction`.

alpine/alpine-infra/src/main/java/alpine/persistence/AlpineQueryManager.java
```diff
@@ -343,6 +343,17 @@ public OidcUser addUserToTeams(final OidcUser user, final List<String> teamNames
...
+    /**
+     * Returns a complete list of all subclasses extending User.class, in ascending order by username.
+     * @return a list of all Users
+     * @since 1.0.0
+     */
+    public List<User> getAllUsers() {
+        final Query<User> query = pm.newQuery(User.class);
+        query.setOrdering("username asc");
+        return executeAndCloseList(query);
+    }
```
nscuro, line 355: Please use [`execute`](https://github.com/DependencyTrack/hyades-apiserver/blob/db00ad60729c0d88e34ab856a62a0459d2aa648b/alpine/alpine-infra/src/main/java/alpine/persistence/AbstractAlpineQueryManager.java#L156-L162) here to make this query support pagination. Since large instances will likely have lots of users, we cannot return them all at once.

apiserver/src/main/java/org/dependencytrack/resources/v1/IntegrationResource.java
```diff
@@ -94,4 +103,34 @@ public Response getInactiveEcosystems() {
...
+        try (QueryManager qm = new QueryManager()) {
+            final ConfigProperty property = qm.getConfigProperty(GITLAB_ENABLED.getGroupName(),
+                    GITLAB_ENABLED.getPropertyName());
+
+            if (!property.getPropertyValue().equals(state)) {
+                if (!state.equalsIgnoreCase("true") && !state.equalsIgnoreCase("false")) {
+                    return Response.status(Response.Status.BAD_REQUEST).build();
+                }
+                property.setPropertyValue(state);
+                qm.persist(property);
```
nscuro, line 129: Please wrap this in a transaction using `qm.callInTransaction`.

Note that `Event.dispatch` must happen *outside* of the transaction. If the TRX has to be rolled back for some reason we don't want the event to be dispatched.

apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
```diff
@@ -360,6 +368,120 @@ public Response uploadBom(@Parameter(required = true) BomSubmitRequest request)
...
+                    cpc.getGroupName(),
+                    cpc.getPropertyName());
+
+            ConfigProperty gitLabIntegrationConfigProperty = propertyGetter.apply(GITLAB_ENABLED);
+            if (gitLabIntegrationConfigProperty == null
+                    || !Boolean.parseBoolean(gitLabIntegrationConfigProperty.getPropertyValue()))
```
nscuro, line 396:
```suggestion
            if (qm.isEnabled(GITLAB_ENABLED))
```

apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
```diff
@@ -360,6 +368,120 @@ public Response uploadBom(@Parameter(required = true) BomSubmitRequest request)
...
+                return Response.notModified("GitLab integration not enabled").build();
+
+            ConfigProperty sbomPushConfigProperty = propertyGetter.apply(GITLAB_SBOM_PUSH_ENABLED);
+            if (sbomPushConfigProperty == null || !Boolean.parseBoolean(sbomPushConfigProperty.getPropertyValue()))
```
nscuro, line 400:
```suggestion
            if (qm.isEnabled(GITLAB_SBOM_PUSH_ENABLED))
```

nscuro (changes requested):
I'd like to see a few minor changes here, but overall looks good. I raise https://github.com/lmco/hyades-apiserver/pull/17 to resolve the merge conflicts.

apiserver/src/main/java/org/dependencytrack/resources/v1/BomResource.java
jhoward-lm: Is this still necessary if the new method performs only read operations? At the end it makes a call to the pre-existing `uploadBom` method, which already uses a transaction

apiserver/src/main/java/org/dependencytrack/integrations/gitlab/GitLabSyncer.java
jhoward-lm: @nscuro I reworked as you suggested. Since the code is in a loop, can you make sure the `qm.runInTransaction` call and `qm.tryAcquireAdvisoryLock` call are happening in the correct place?

PR #1462, opened 2025-10-02
Support for CSAF

oxisto:
### Description

This PR adds support for CSAF as a discovery source and introduces a new concept called "advisories".

See https://github.com/DependencyTrack/hyades/pull/1894 for the vulnerability analyser in the Hyades repo

### Addressed Issue

Addresses https://github.com/DependencyTrack/dependency-track/issues/2353

### Checklist

- [x] ~This PR fixes a defect, and I have provided tests to verify that the fix is effective~
- [x] ~This PR introduces new or alters existing behavior, and I have updated the [documentation] accordingly~

### DIFF
--- a/persistence-migration/src/main/resources/migration/changelog-v5.7.0.xml
+++ b/persistence-migration/src/main/resources/migration/changelog-v5.7.0.xml
@@ -359,44 +359,8 @@
         </createTable>
     </changeSet>
 
+
     <changeSet id="v5.7.0-18" author="oxisto">
-        <sql splitStatements="false">
-            DO $$
-            DECLARE
-            mat_view_def TEXT;
-            index_defs TEXT[];
-            index_def TEXT;
-            BEGIN
-            SELECT pg_get_viewdef('"PORTFOLIOMETRICS_GLOBAL"', true)
-            INTO mat_view_def;
-
-            SELECT array_agg(indexdef)
-            INTO index_defs
-            FROM pg_indexes
-            WHERE tablename = 'PORTFOLIOMETRICS_GLOBAL';
-
-            DROP MATERIALIZED VIEW "PORTFOLIOMETRICS_GLOBAL";
-
-            ALTER TABLE "CONFIGPROPERTY"
-            ALTER COLUMN "PROPERTYVALUE" TYPE TEXT;
-
-            EXECUTE format(
-            'CREATE MATERIALIZED VIEW "PORTFOLIOMETRICS_GLOBAL" AS %s WITH DATA'
-            , TRIM(TRAILING ';' FROM mat_view_def)
-            );
-
-            IF index_defs IS NOT NULL THEN
-            FOREACH index_def IN ARRAY index_defs
-            LOOP
-            EXECUTE index_def;
-            END LOOP;
-            END IF;
-            END
-            $$;
-        </sql>
-    </changeSet>
-
-    <changeSet id="v5.7.0-19" author="oxisto">
         <createTable tableName="ADVISORY">
             <column name="ID" autoIncrement="true" type="BIGINT">
                 <constraints nullable="false" primaryKey="true" primaryKeyName="ADVISORY_PK"/>
@@ -423,7 +387,7 @@
         </createTable>
     </changeSet>
 
-    <changeSet id="v5.7.0-20" author="oxisto">
+    <changeSet id="v5.7.0-19" author="oxisto">
         <sql>
             ALTER TABLE "ADVISORY"
             ADD COLUMN "SEARCHVECTOR" tsvector
@@ -435,7 +399,7 @@
         </sql>
     </changeSet>
 
-    <changeSet id="v5.7.0-21" author="oxisto">
+    <changeSet id="v5.7.0-20" author="oxisto">
         <addColumn tableName="VULNERABILITYALIAS">
             <column name="CSAF_ID" type="VARCHAR(255)"/>
         </addColumn>
@@ -443,12 +407,12 @@
             <column name="CSAF_ID"/>
         </createIndex>
     </changeSet>
-    <changeSet id="v5.7.0-22" author="oxisto">
+    <changeSet id="v5.7.0-21" author="oxisto">
         <addColumn tableName="FINDINGATTRIBUTION">
             <column name="MATCHING_PERCENTAGE" type="SMALLINT"/>
         </addColumn>
     </changeSet>
-    <changeSet id="v5.7.0-23" author="oxisto">
+    <changeSet id="v5.7.0-22" author="oxisto">
         <createTable tableName="ADVISORIES_VULNERABILITIES">
             <column name="ADVISORY_ID" type="BIGINT">
                 <constraints nullable="false"/>


plugin/vuln-data-source-csaf/src/main/java/org/dependencytrack/datasource/vuln/csaf/CsafVulnDataSourceConfigs.java
```diff
@@ -0,0 +1,85 @@
...
+        try {
+            defaultApiUrl = URI.create("https://wid.cert-bund.de/.well-known/csaf-aggregator/aggregator.json").toURL();
+        } catch (MalformedURLException e) {
+            throw new IllegalStateException("Failed to parse default API URL", e);
+        }
+
+        final ObjectMapper objectMapper = new ObjectMapper().registerModule(new JavaTimeModule());;
+        final String defaultSources;
+        try {
+            defaultSources = objectMapper.writeValueAsString(new ArrayList<>(List.of(
```
nscuro: Since all values are known in advance, could this be a hardcoded JSON string instead?

apiserver/src/main/java/org/dependencytrack/model/ConfigPropertyConstants.java
```diff
@@ -66,7 +66,10 @@ public enum ConfigPropertyConstants {
...
     SCANNER_SNYK_CVSS_SOURCE("scanner", "snyk.cvss.source", "NVD", PropertyType.STRING, "Type of source to be prioritized for cvss calculation", ConfigPropertyAccessMode.READ_WRITE),
     SCANNER_SNYK_BASE_URL("scanner", "snyk.base.url", "https://api.snyk.io", PropertyType.URL, "Base Url pointing to the hostname and path for Snyk analysis", ConfigPropertyAccessMode.READ_WRITE),
+    SCANNER_CSAF_ENABLED("scanner", "csaf.enabled", "false", PropertyType.BOOLEAN, "Flag to enable/disable CSAF Vulnerability Analysis", ConfigPropertyAccessMode.READ_WRITE),
+    SCANNER_CSAF_THRESHOLD("scanner", "csaf.threshold", "75", PropertyType.INTEGER, "Matching threshold for CSAF Vulnerability Analysis", ConfigPropertyAccessMode.READ_WRITE),
     VULNERABILITY_POLICY_FILE_LAST_MODIFIED_HASH("vulnerability-policy", "vulnerability.policy.file.last.modified.hash", null,  PropertyType.STRING, "Hash value or etag of the last fetched bundle if any", ConfigPropertyAccessMode.READ_ONLY),
+    VULNERABILITY_SOURCE_CSAF_ENABLED("vuln-source", "csaf.enabled", "false", PropertyType.BOOLEAN, "Flag to enable/disable CSAF Advisories", ConfigPropertyAccessMode.READ_WRITE),
```
nscuro: Should use `VulnDataSourceFactory#isDataSourceEnabled()` instead. A separate `enabled` property should not be necessary.

apiserver/src/main/java/org/dependencytrack/resources/v1/AdvisoriesResource.java
```diff
@@ -0,0 +1,380 @@
...
+import static org.dependencytrack.persistence.jdbi.JdbiFactory.withJdbiHandle;
+import static org.dependencytrack.resources.v1.FindingResource.MEDIA_TYPE_SARIF_JSON;
+
+/**
+ * JAX-RS resources for advisories.
+ *
+ * @author Lawrence Dean
+ * @since TODO set version
+ */
+@Path("/v1/advisories")
```
nscuro: v1 of the API uses the singular form of nouns. In [v2](https://github.com/DependencyTrack/hyades-apiserver/blob/main/api/src/main/openapi/openapi.yaml) we have adopted the (more common) plural form.

On that note, it would be ideal if new endpoints were added to v2, as it makes quality control easier with linting and breaking change detection etc.

apiserver/src/main/java/org/dependencytrack/resources/v1/AdvisoriesResource.java
```diff
@@ -0,0 +1,380 @@
...
+            LOGGER.debug("Synchronizing vulnerability " + vuln.getVulnId());
+            final Vulnerability persistentVuln = qm.synchronizeVulnerability(vuln, false);
+            qm.synchronizeVulnerableSoftware(persistentVuln, vsList, Vulnerability.Source.CSAF);
+
+            advisory.addVulnerability(persistentVuln);
+        }
+
+        // Sync the document into the database, replacing an older version with the same
+        // combination of tracking ID and publisher namespace if necessary
+        qm.synchronizeAdvisory(advisory);
```
nscuro: When multiple persistence operations are performed, please wrap them in a `qm.runInTransaction` or `qm.callInTransaction` block to ensure atomicity.

apiserver/src/main/java/org/dependencytrack/resources/v1/AdvisoriesResource.java
```diff
@@ -0,0 +1,380 @@
...
+            @ApiResponse(responseCode = "204", description = "Advisory removed successfully"),
+            @ApiResponse(responseCode = "401", description = "Unauthorized"),
+            @ApiResponse(responseCode = "404", description = "The advisory could not be found")
+    })
+    @PermissionRequired(Permissions.Constants.VULNERABILITY_ANALYSIS_UPDATE)
+    public Response deleteAdvisory(@PathParam("advisoryId") String advisoryId) {
+        try (QueryManager qm = new QueryManager()) {
+            final Advisory entity = qm.getObjectById(Advisory.class, advisoryId);
+            if (entity != null) {
+                qm.delete(entity);
```
nscuro: When multiple persistence operations are performed, please wrap them in a `qm.runInTransaction` or `qm.callInTransaction` block to ensure atomicity.

apiserver/src/main/java/org/dependencytrack/resources/v1/AdvisoriesResource.java
```diff
@@ -0,0 +1,380 @@
...
+            @ApiResponse(responseCode = "401", description = "Unauthorized"),
+            @ApiResponse(responseCode = "404", description = "The advisory could not be found")
+    })
+    @PermissionRequired(Permissions.Constants.VULNERABILITY_ANALYSIS_UPDATE)
+    public Response markAdvisoryAsSeen(@PathParam("advisoryId") String advisoryId) {
+        try (QueryManager qm = new QueryManager()) {
+            final Advisory entity = qm.getObjectById(Advisory.class, advisoryId);
+            if (entity != null) {
+                entity.setSeen(true);
+                qm.persist(entity);
```
nscuro: When multiple persistence operations are performed, please wrap them in a `qm.runInTransaction` or `qm.callInTransaction` block to ensure atomicity.

apiserver/src/main/java/org/dependencytrack/resources/v1/AdvisoriesResource.java
```diff
@@ -0,0 +1,380 @@
...
+                        ? numAffectedComponentsBoxed
+                        : 0L;
+
+                return Response.ok(new AdvisoryDao.AdvisoryResult(
+                        advisoryEntity,
+                        affectedProjects,
+                        numAffectedComponents,
+                        vulnerabilities)).build();
+            }
+        }
```
nscuro: Use either the JDO API (`QueryManager`), *or* JDBI, but don't mix them in the same block. They cannot share transactions or connections. Keeping a handle of both of them open wastes resources.

apiserver/src/main/java/org/dependencytrack/resources/v1/AdvisoriesResource.java
```diff
@@ -0,0 +1,380 @@
...
+            LOGGER.info("Querying for " + projectId + " :: " + advisoryId);
+            List<AdvisoryDao.ProjectAdvisoryFindingRow> advisoryRows = withJdbiHandle(getAlpineRequest(),
+                    handle -> handle.attach(AdvisoryDao.class)
+                            .getFindingsByProjectAdvisory(projectId, advisoryId));
+            final long totalCount = advisoryRows.size();
+            LOGGER.info("retrieved size " + totalCount);
+
+            return Response.ok(advisoryRows.stream().toList()).header(TOTAL_COUNT_HEADER, totalCount)
+                    .build();
+        }
```
nscuro: Use either the JDO API (`QueryManager`), *or* JDBI, but don't mix them in the same block. They cannot share transactions or connections. Keeping a handle of both of them open wastes resources.

apiserver/src/main/java/org/dependencytrack/util/CsafUtil.java
```diff
@@ -0,0 +1,38 @@
...
+
+public final class CsafUtil {
+    private CsafUtil() {}
+
+    public static boolean validateUrl(String url) {
+        return UrlValidator.getInstance().isValid(url);
+    }
+
+    public static boolean validateDomain(String domain) {
+        return DomainValidator.getInstance().isValid(domain);
```
nscuro: Can we find a way to do this without requiring an additional dependency?

dev/scripts/dbschema-generate.sh
```diff
@@ -22,7 +22,7 @@ set -euo pipefail
 SCRIPT_DIR="$(cd -P -- "$(dirname "$0")" && pwd -P)"
 ROOT_DIR="$(cd -P -- "${SCRIPT_DIR}/../../" && pwd -P)"
 MIGRATION_DIR="$(cd -P -- "${ROOT_DIR}/persistence-migration" && pwd -P)"
-CONTAINER_ID="$(docker run -d --rm -e 'POSTGRES_DB=dtrack' -e 'POSTGRES_USER=dtrack' -e 'POSTGRES_PASSWORD=dtrack' -p '5432' postgres:13-alpine)"
+CONTAINER_ID="$(docker run -d --rm -e 'POSTGRES_DB=dtrack' -e 'POSTGRES_USER=dtrack' -e 'POSTGRES_PASSWORD=dtrack' -p '5432' postgres:17-alpine)"
```
nscuro: 13 is our current lowest supported version. All tests must work with it. Do you need a PG feature that is only available in newer versions?

nscuro (changes requested):

dev/scripts/dbschema-generate-migration.sh
```diff
@@ -22,7 +22,7 @@ set -euo pipefail
 SCRIPT_DIR="$(cd -P -- "$(dirname "$0")" && pwd -P)"
 ROOT_DIR="$(cd -P -- "${SCRIPT_DIR}/../../" && pwd -P)"
 MIGRATION_DIR="$(cd -P -- "${ROOT_DIR}/persistence-migration" && pwd -P)"
-CONTAINER_ID="$(docker run -d --rm -e 'POSTGRES_DB=dtrack' -e 'POSTGRES_USER=dtrack' -e 'POSTGRES_PASSWORD=dtrack' -p '5432' postgres:13-alpine)"
+CONTAINER_ID="$(docker run -d --rm -e 'POSTGRES_DB=dtrack' -e 'POSTGRES_USER=dtrack' -e 'POSTGRES_PASSWORD=dtrack' -p '5432' postgres:17-alpine)"
```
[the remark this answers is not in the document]
oxisto: So it seems that all features should be available in 13, but just FYI Hyades also uses 17 in its docker-compose https://github.com/DependencyTrack/hyades/blob/8fffd989e3487d0c589ef1eb580facc6a0064c91/docker-compose.yml#L170 so there might be a good idea to synchronise this at some point

apiserver/src/main/java/org/dependencytrack/util/CsafUtil.java
oxisto: I did a very crude implementation of a domain validator and also moved it to the `.validation` package.

plugin/vuln-data-source-csaf/src/main/java/org/dependencytrack/datasource/vuln/csaf/CsafVulnDataSource.java
```diff
@@ -0,0 +1,285 @@
...
+import static org.dependencytrack.datasource.vuln.csaf.CycloneDxPropertyNames.PROPERTY_ADVISORY_PROVIDER_ID;
+import static org.dependencytrack.datasource.vuln.csaf.CycloneDxPropertyNames.PROPERTY_ADVISORY_UPDATED;
+import static org.dependencytrack.datasource.vuln.csaf.CycloneDxPropertyNames.extractProperty;
+
+/**
+ * A Vulnerability Data Source that retrieves and processes CSAF documents from configured sources.
+ *
+ * @since 5.7.0
+ */
+public class CsafVulnDataSource implements VulnDataSource {
```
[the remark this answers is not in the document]
oxisto: Hmm probably not that easy without spinning up a HTTP server. Not sure if we can easily mock the return values of the kotlin-csaf libraries out of dependencytrack.

apiserver/pom.xml
```diff
@@ -114,6 +114,11 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.dependencytrack</groupId>
+            <artifactId>vuln-data-source-csaf</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
nscuro: The transitive `slf4j-jdk14` dependency seems to conflict with our desired Logback slf4j provider:

```suggestion
            <exclusions>
                <exclusion>
                    <groupId>org.slf4j</groupId>
                    <artifactId>slf4j-jdk14</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
```

persistence-migration/src/main/resources/migration/changelog-v5.7.0.xml
```diff
@@ -250,4 +250,80 @@
...
+        <addColumn tableName="VULNERABILITYALIAS">
+            <column name="CSAF_ID" type="VARCHAR(255)"/>
+        </addColumn>
+        <createIndex indexName="VULNERABILITYALIAS_CSAF_ID_IDX" tableName="VULNERABILITYALIAS">
+            <column name="CSAF_ID"/>
+        </createIndex>
+    </changeSet>
+    <changeSet id="v5.7.0-15" author="oxisto">
+        <addColumn tableName="FINDINGATTRIBUTION">
+            <column name="MATCHING_PERCENTAGE" type="FLOAT"/>
```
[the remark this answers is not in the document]
nscuro: Since this stores values from `0.00` to `1.00`:
  * Either `NUMERIC(3,2)`, meaning 3 total digits and 2 after decimal point.
  * Or, even simpler: Use `SMALLINT` and store the percentage as values from 0 to 100.

The latter would also prevent any rounding issues when converting between the database type and the Java type.

apiserver/src/main/java/org/dependencytrack/model/Advisory.java
```diff
@@ -0,0 +1,249 @@
...
+@PersistenceCapable
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class Advisory implements Serializable {
+
+    /**
+     * The internal database id of the advisory.
+     */
+    @PrimaryKey
+    @Persistent(valueStrategy = IdGeneratorStrategy.NATIVE)
+    private long id;
```
nscuro: Internal IDs should not be disclosed by the API by accident:

```suggestion
    @JsonIgnore
    private long id;
```

apiserver/src/main/java/org/dependencytrack/resources/v2/AdvisoriesResource.java
```diff
@@ -0,0 +1,365 @@
...
+    @Produces({ MediaType.APPLICATION_JSON, MEDIA_TYPE_SARIF_JSON })
+    @Operation(summary = "Returns a list of all advisories", description = "<p>Requires permission <strong>VIEW_VULNERABILITY</strong></p>")
+    @ApiResponses(value = {
+            @ApiResponse(responseCode = "200", description = "A list of all advisories for a specific project, or a SARIF file", headers = @Header(name = TOTAL_COUNT_HEADER, description = "The total number of advisories", schema = @Schema(format = "integer")), content = {
+                    @Content(array = @ArraySchema(schema = @Schema(implementation = AdvisoryDao.AdvisoryInProjectRow.class)), mediaType = MediaType.APPLICATION_JSON),
+                    @Content(schema = @Schema(type = "string"), mediaType = MEDIA_TYPE_SARIF_JSON)
+            }),
+            @ApiResponse(responseCode = "401", description = "Unauthorized"),
+            @ApiResponse(responseCode = "403", description = "Access to advisories prohibited", content = @Content(schema = @Schema(implementation = ProblemDetails.class), mediaType = ProblemDetails.MEDIA_TYPE_JSON)),
+    })
```
nscuro: Does not return SARIF.

```suggestion
    @GET
    @Produces(MediaType.APPLICATION_JSON)
    @Operation(summary = "Returns a list of all advisories", description = "<p>Requires permission <strong>VIEW_VULNERABILITY</strong></p>")
    @ApiResponses(value = {
            @ApiResponse(responseCode = "200", description = "A list of all advisories for a specific project", headers = @Header(name = TOTAL_COUNT_HEADER, description = "The total number of advisories", schema = @Schema(format = "integer")), content = @Content(array = @ArraySchema(schema = @Schema(implementation = AdvisoryDao.AdvisoryInProjectRow.class)), mediaType = MediaType.APPLICATION_JSON)),
            @ApiResponse(responseCode = "401", description = "Unauthorized"),
            @ApiResponse(responseCode = "403", description = "Access to advisories prohibited", content = @Content(schema = @Schema(implementation = ProblemDetails.class), mediaType = ProblemDetails.MEDIA_TYPE_JSON)),
    })
```

plugin/vuln-data-source-csaf/src/main/java/org/dependencytrack/datasource/vuln/csaf/CsafVulnDataSourceConfigs.java
oxisto: I‘d rather leave it as is then

apiserver/pom.xml
oxisto: That’s why I got these weird log error messages.

persistence-migration/src/main/resources/migration/changelog-v5.7.0.xml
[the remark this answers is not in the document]
oxisto: I used `SMALLINT` now, even though we have to do some int/short converting in the java code, which should be fine since we are not exceeding 100 anyways.

oxisto:
Now the NVD tests fail because the feed is not available :(

apiserver/src/main/java/org/dependencytrack/resources/v2/AdvisoriesResource.java
```diff
@@ -0,0 +1,364 @@
...
+    public Response deleteAdvisory(String advisoryId) {
+        try (final var qm = new QueryManager()) {
+            return qm.callInTransaction(() -> {
+                final Advisory entity = qm.getObjectById(Advisory.class, advisoryId);
+                if (entity != null) {
+                    qm.delete(entity);
+                    return Response.status(Response.Status.NO_CONTENT).build();
+                } else {
+                    return Response.status(Response.Status.NOT_FOUND)
+                            .entity("The advisory could not be found.").build();
```
nscuro:
```suggestion
                    throw new NotFoundException();
```

apiserver/src/main/java/org/dependencytrack/resources/v2/CsafResource.java
```diff
@@ -0,0 +1,284 @@
...
+
+    @Override
+    @PermissionRequired(Permissions.Constants.VULNERABILITY_MANAGEMENT_DELETE)
+    public Response deleteCsafSource(Integer csafSourceId) {
+        // Fetch existing sources and look for the one to delete
+        var sources = new ArrayList<>(getPluginCsafSourcesFromConfig(null));
+        var source = getCsafSourceByIdFromConfig(sources, csafSourceId);
+        if (source == null) {
+            return Response.status(Response.Status.NOT_FOUND)
+                    .entity("The ID of the source could not be found.").build();
```
nscuro:
```suggestion
            throw new NotFoundException();
```

nscuro:
> But somehow I cannot get this fork running for me at all now locally. I cannot login with the default admin / admin anymore, which should trigger the initial database-reset

Try these things:

* Does this also occur in dev mode (`mvn -Pquick -Pdev-services -pl apiserver -am verify`)?
  * It doesn't for me.
* If you're connecting to a DB started separately:
  * Check the `databasechangelog` table: Are all DB migrations recorded there?
  * Is it an option for you to nuke the DB and start from a fresh one?
* Possible the Maven build cache got corrupted. Try nuking it: `rm -rf ~/.m2/build-cache/v1.1/org.dependencytrack/`.

> Now the NVD tests fail because the feed is not available :(

Admittedly I was a bit lazy in using the actual NVD endpoint for this test. I'll work on replacing it with WireMock instead to avoid this being a source of flakiness.

nscuro (commented):
Thanks for taking the time to adapt the API endpoints to v2, it's really much appreciated!

Some minor nits on this. In general, we want to make all errors return [RFC9457 Problem Details](https://www.rfc-editor.org/rfc/rfc9457.html) so they're more easily machine-readable. This is already handled for you by JAX-RS exception mappers, it's only a matter of throwing the correct exceptions.

We also have linting to ensure the v2 API stays consistent. For some reason the linting workflow passed despite having identified issues. Please have a look at the findings by running the linting script manually while we fix the workflow.

nscuro:
Seems like these migrations got lost in a recent merge from `main`: https://github.com/DependencyTrack/hyades-apiserver/blob/0630a7d446061722be1d262a4edafd8bb1e2da70/persistence-migration/src/main/resources/migration/changelog-v5.7.0.xml#L254-L316

oxisto:
> @oxisto Can you commit what you have tried? Happy to have a look what the underlying issue is.

thanks! Pushed it. 

Somehow the development environment is very unstable or my mvn skills are inferior, every 2 days I seem to completely destroy my environment and cannot execute any tests, with missing dependency to code-coverage and other internal projects :(

nscuro:
@oxisto If you can share a few more details as to how you're building the project / run tests I'll see how we can improve the setup.

Generally though:

* You need to run all `mvn` commands from the project root, otherwise Maven won't correctly resolve inter-module dependencies. I appreciate this might be different to what Gradle allows for.
* To focus on individual submodules you should use the `-pl` flag, e.g. `mvn -pl apiserver`.
* To make Maven build all modules `apiserver` depends on, pass the `-am` (`--also-make`) flag.
* e.g. `mvn -pl apiserver -am clean package -Pquick`

A less involved solution is to run `mvn clean install -Pquick`, which will install all modules to the local Maven repository. The build cache should make this operation fairly fast.

apiserver/src/main/java/org/dependencytrack/resources/v2/AdvisoriesResource.java
```diff
@@ -0,0 +1,375 @@
...
+
+            return Response.ok(response).header(TOTAL_COUNT_HEADER, totalCount).build();
+        });
+    }
+
+    @Override
+    @PermissionRequired(Permissions.Constants.VULNERABILITY_ANALYSIS_CREATE)
+    public Response uploadAdvisory(
+            @QueryParam("format") @NotNull String format,
+            @FormDataParam("file") InputStream _fileInputStream) {
```
nscuro: Not exactly sure why, but these annotations somehow cause the method definition in `AdvisoryApi` to be shadowed. And since this particular method doesn't have a `@Path`, Jersey doesn't register it as valid resource method at all.

Removing the annotations makes the tests pass.

```suggestion
            String format,
            InputStream _fileInputStream) {
```

nscuro (approved):

PR #1867, opened 2026-03-11
Improve component search performance with JDBI

### LINKED ISSUE #2049: Improve performance of component search by coordinates
### Current Behavior

The component search query based on coordinates (group, name, version) is slow and inefficient (takes ~12s for fetching over 500 records) which is resulting into `EoFException`.
There are indices present on the columns but currently project association for each component likely causes N+1 queries.

### Steps to Reproduce

1. Call the API endpoint `/api/v1/component/identity?name=componentName&version=version` OR query the database to search for a component based on group/name/version.
2. The query takes too long returning `504 Gateway Time-Out`.

### Expected Behavior

The query should return paginated result of all the components matching the search criteria.

**Solution**: Convert the SQL to single and optimised JDBI query.

### Repository Type

Hyades apiserver

### Checklist

- [x] I have read and understand the [contributing guidelines](https://github.com/DependencyTrack/hyades/blob/master/CONTRIBUTING.md#filing-issues)
- [x] I have checked the [existing issues](https://github.com/DependencyTrack/hyades/issues) for whether this defect was already reported

sahibamittal:
### Description

1. This change converts the SQL to single and optimised JDBI query for searching component by identity.
2. Introduces new API endpoint to search components by identity under v2.
Old and existing: `/api/v1/component/identity`
New: `/api/v2/components`

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/2049

### Additional Details

Frontend changes: https://github.com/DependencyTrack/hyades-frontend/pull/438

### Checklist


api/src/main/openapi/components/schemas/project.yaml
nscuro: This is a limited view on the project resource, presumably used in a specific context. The name of the schema should reflect that to avoid confusion, i.e. the next time an endpoint needs to represent a "project", it likely shouldn't reuse this one.

nscuro (changes requested):

nscuro (changes requested):

api/src/main/openapi/components/schemas/component-project.yaml
```diff
@@ -0,0 +1,29 @@
...
+    type: string
+    maxLength: 255
+  version:
+    type: string
+    maxLength: 255
+  uuid:
+    type: string
+    format: uuid
+  direct_dependencies:
+    type: string
```
nscuro: Looking at the frontend, it seems this is not *really* needed: https://github.com/DependencyTrack/hyades-frontend/pull/438/changes#diff-8d6a992f66aad71ba9f794db71383e8816847ad7c7f4d10cc15a2cd55fce22adR214

Its presence is used as an indicator for whether the project has a dependency graph at all, so that the "show in dependency graph" button is rendered conditionally.

I'd argue we should omit this field, and render the button in the frontend unconditionally. If no graph is available, the frontend can simply show a warning or an error when the button is clicked.

nscuro (changes requested):

apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ComponentDao.java
```diff
@@ -193,15 +206,291 @@ List<Component> listProjectComponents(
...
+                                <#if sortByColumn == "lastInheritedRiskScore" > CAST(:lastPrimaryValue AS DOUBLE PRECISION)
+                                <#else>:lastPrimaryValue
+                                </#if>
+                             AND "C"."ID" > :lastId))
+                        </#if>
+                    )
+                <#elseif hasCursor && lastPrimaryValue?has_content && lastSecondaryValue?has_content && lastId?has_content>
+                    AND ("C"."NAME" > :lastPrimaryValue
+                            OR ("C"."NAME" = :lastPrimaryValue AND "C"."VERSION" < :lastSecondaryValue)
+                            OR ("C"."NAME" = :lastPrimaryValue AND "C"."VERSION" = :lastSecondaryValue AND "C"."ID" > :lastId))
```
nscuro: `VERSION` column can be removed here since the default sorting no longer includes it.

apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ComponentDao.java
```diff
@@ -193,15 +206,291 @@ List<Component> listProjectComponents(
...
+                    totalCount);
+        } else {
+            nextPageToken = null;
+        }
+
+        if (includeMetrics) {
+            final Map<Long, Component> componentById = resultRows.stream()
+                    .collect(Collectors.toMap(Component::getId, Function.identity()));
+            final List<DependencyMetrics> metricsList = withJdbiHandle(
+                    handle -> handle.attach(MetricsDao.class).getMostRecentDependencyMetrics(componentById.keySet()));
```
nscuro: Use `getHandle()` here instead of creating a new nested handle (`withJdbiHandle`).

apiserver/src/main/java/org/dependencytrack/persistence/jdbi/ComponentDao.java
```diff
@@ -193,15 +206,291 @@ List<Component> listProjectComponents(
...
+                                <#else> :lastPrimaryValue
+                                </#if>
+                             OR ("C"."${sortByColumn}" =
+                                <#if sortByColumn == "lastInheritedRiskScore" > CAST(:lastPrimaryValue AS DOUBLE PRECISION)
+                                <#else>:lastPrimaryValue
+                                </#if>
+                             AND "C"."ID" > :lastId))
+                        </#if>
+                    )
+                <#elseif hasCursor && lastPrimaryValue?has_content && lastSecondaryValue?has_content && lastId?has_content>
```
nscuro: When sorting by `last_inherited_risk_score`, advancing to the second page fails with:

```
org.postgresql.util.PSQLException: ERROR: column C.lastInheritedRiskScore does not exist
```

The problem appears to be that `lastInheritedRiskScore` is a column alias, but it's always appended to `"C"` here which is not valid. Either it needs to be just `"lastInheritedRiskScore"`, *or* `"C"."RISKSCORE"`.

nscuro (changes requested):

nscuro (approved):

PR #1944, opened 2026-03-27
added changes to fix affected project search in vulnerabilities page

mehab:
### Description
This PR fixes the affected project search on the vulnerability detail page. Currently this search does not refresh if project name is provided in the search bar

### Addressed Issue
If the number of affected projects is too many for a vulnerability, the search box should help us search specific project as needed. Currently even if we type something in the search box, the list does not get refreshed.

### Checklist


apiserver/src/main/java/org/dependencytrack/resources/v1/VulnerabilityResource.java
```diff
@@ -260,7 +262,22 @@ public Response getVulnerabilityByVulnId(
     @Produces(MediaType.APPLICATION_JSON)
     @Operation(
             summary = "Returns a list of all projects affected by a specific vulnerability",
-            description = "<p>Requires permission <strong>VIEW_PORTFOLIO</strong></p>"
+            description = """
+                    <p>Requires permission <strong>VIEW_PORTFOLIO</strong></p>\
+                    <p>Optional query parameters <code>searchText</code> or <code>filter</code> narrow the list; \
+                    both are provided by the Alpine request filter and match the same value.</p>""",
```
nscuro: This is an implementation detail that doesn't belong in public API documentation.

apiserver/src/main/java/org/dependencytrack/resources/v1/VulnerabilityResource.java
```diff
@@ -260,7 +262,22 @@ public Response getVulnerabilityByVulnId(
     @Produces(MediaType.APPLICATION_JSON)
     @Operation(
             summary = "Returns a list of all projects affected by a specific vulnerability",
-            description = "<p>Requires permission <strong>VIEW_PORTFOLIO</strong></p>"
+            description = """
+                    <p>Requires permission <strong>VIEW_PORTFOLIO</strong></p>\
+                    <p>Optional query parameters <code>searchText</code> or <code>filter</code> narrow the list; \
```
nscuro: Let's commit to only `searchText` to avoid confusion. Also, when we explicitly mention it, we should also note *what* it searches (i.e. project name, case-insensitive, "contains" semantics).

apiserver/src/main/java/org/dependencytrack/resources/v1/VulnerabilityResource.java
```diff
@@ -278,8 +295,16 @@ public Response getAffectedProject(@PathParam("source") String source,
                                        @PathParam("vuln") String vuln,
                                        @Parameter(description = "Optionally excludes inactive projects from being returned", required = false)
                                        @QueryParam("excludeInactive") boolean excludeInactive) {
-        final List<AffectedProjectListRow> affectedProjectRows = withJdbiHandle(getAlpineRequest(), handle ->
-                handle.attach(VulnerabilityDao.class).getAffectedProjects(source, vuln, excludeInactive ? true : null));
+        final AlpineRequest alpineRequest = getAlpineRequest();
+        final String affectedProjectsFilter;
+        if (alpineRequest == null) {
```
nscuro: `getAlpineRequest` should never return `null`, no need to handle that here.

nscuro (changes requested):
Let's get the API docs cleaned up, otherwise looks good.

nscuro (dismissed):

nscuro:
@mehab The new tests are failing, I think you need to add `initializeWithPermissions` so the requests are authorized correctly.

nscuro (approved):

PR #1979, opened 2026-04-16
Add Metrics update trigger after cloning a project

ElenaStroebele:
### Description
Ports [https://github.com/DependencyTrack/dependency-track/pull/4806 ](url).

### Addressed Issue
Relates to [https://github.com/DependencyTrack/hyades/issues/2105](url).

### Additional Details
Changed the logic a little bit, due to different event handling in Hyades.

### Checklist

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/resources/v2/ProjectsResource.java
+++ b/apiserver/src/main/java/org/dependencytrack/resources/v2/ProjectsResource.java
@@ -52,7 +52,6 @@ import java.util.List;
 import java.util.UUID;
 
 import static org.dependencytrack.persistence.jdbi.JdbiFactory.inJdbiTransaction;
-import static org.dependencytrack.persistence.jdbi.JdbiFactory.useJdbiTransaction;
 import static org.dependencytrack.resources.v2.mapping.ModelMapper.mapHashes;
 import static org.dependencytrack.resources.v2.mapping.ModelMapper.mapLicense;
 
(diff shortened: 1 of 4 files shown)


apiserver/src/main/java/org/dependencytrack/resources/v2/ProjectsResource.java
```diff
@@ -132,6 +134,11 @@ public Response cloneProject(final UUID projectUuid, final CloneProjectRequest r
                             request.getIncludes().contains(CloneProjectInclude.TAGS)));
         });
 
+        useJdbiTransaction(getAlpineRequest(), handle -> {
```
nscuro: Let's do the metrics update in the same transaction that does the cloning to keep things consistent.

apiserver/src/main/java/org/dependencytrack/tasks/CloneProjectTask.java
```diff
@@ -84,6 +86,9 @@ public void inform(final Event e) {
                                             request.includeServices(),
                                             request.includeTags())));
 
+                    useJdbiTransaction(handle ->
```
nscuro: Let's do the metrics update in the same transaction that does the cloning to keep things consistent.

nscuro (changes requested):
Two minor comments WRT transaction scoping, otherwise looks good!

apiserver/src/main/java/org/dependencytrack/resources/v2/ProjectsResource.java
ElenaStroebele: Implemented the requested change!

apiserver/src/main/java/org/dependencytrack/tasks/CloneProjectTask.java
ElenaStroebele: Implemented the requested change!

nscuro (approved):

PR #1989, opened 2026-04-16
Port : Enable source filtering in SARIF format for `/finding/project/{UUID}`

sahibamittal:
### Description

Enable missing source filtering for SARIF format in resource `\finding\project\{uuid}`

### Addressed Issue

Relates to https://github.com/DependencyTrack/hyades/issues/2105
Ports https://github.com/DependencyTrack/dependency-track/pull/4949

### Checklist

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/resources/v1/FindingResource.java
+++ b/apiserver/src/main/java/org/dependencytrack/resources/v1/FindingResource.java
@@ -153,6 +153,9 @@ public class FindingResource extends AbstractApiResource {
                         handle.attach(FindingDao.class).getFindingsByProject(project.getId(), /* includeInactive */ false, suppressed, hasAnalysis));
                 final long totalCount = findingRows.isEmpty() ? 0 : findingRows.getFirst().totalCount();
                 List<Finding> findings = findingRows.stream().map(Finding::new).toList();
+                if (source != null) {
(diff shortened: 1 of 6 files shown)


apiserver/src/main/java/org/dependencytrack/resources/v1/FindingResource.java
```diff
@@ -153,6 +153,9 @@ public Response getFindingsByProject(@Parameter(description = "The UUID of the p
                         handle.attach(FindingDao.class).getFindingsByProject(project.getId(), /* includeInactive */ false, suppressed, hasAnalysis));
                 final long totalCount = findingRows.isEmpty() ? 0 : findingRows.getFirst().totalCount();
                 List<Finding> findings = findingRows.stream().map(Finding::new).toList();
+                if (source != null) {
+                    findings = findings.stream().filter(finding -> source.equals(finding.getVulnerability().get("source"))).collect(Collectors.toList());
+                }
                 findings = mapComponentLatestVersion(findings);
```
nscuro: I think this filter no longer works as expected, and was broken even before. The issue is that in v4, this endpoint was not paginated, so filtering in-memory was fine. In v5 we added pagination, so filtering in-memory breaks pagination. The filtering would now need to happen at the query level.

nscuro (approved):

PR #1122, opened 2025-04-07
Add `severity` as postgres enum `type`

sahibamittal:
### Description

Make `SEVERITY` a PostgreSQL enum.
This is present as column in tables `VULNERABILITY` and `ANALYSIS`.

### Addressed Issue

Closes https://github.com/DependencyTrack/hyades/issues/1725

### Checklist

- [x] I have read and understand the [contributing guidelines]
- [ ] This PR fixes a defect, and I have provided tests to verify that the fix is effective
- [x] This PR implements an enhancement, and I have provided tests to verify that it works as intended
- [ ] This PR introduces changes to the database model, and I have updated the [migration changelog] accordingly
- [ ] This PR introduces new or alters existing behavior, and I have updated the [documentation] accordingly

[contributing guidelines]: ../CONTRIBUTING.md#pull-requests
[documentation]: https://dependencytrack.github.io/hyades/latest/development/documentation/
[migration changelog]: https://dependencytrack.github.io/hyades/latest/development/database-migrations/

PR #1991, opened 2026-04-16
Fix findings API returning 500 when filtering by severity

ider-criteo:
### Description

`GET /api/v1/finding?severity=CRITICAL` returns 500 with `operator does not exist: severity = character varying`. This regression was introduce with the migration done in #1122. It converted the `SEVERITY` column to an enum, but `FindingDao.processArrayFilter` still binds filter values as plain varchar strings, making the API fail.
Casting the column to `TEXT` before comparison in `processArrayFilter` makes the function work for both enum and varchar columns.

### Checklist

- [ ] ~This PR implements an enhancement, and I have provided tests to verify that it works as intended~
- [ ] ~This PR introduces new or alters existing behavior, and I have updated the [documentation] accordingly~

### DIFF
--- a/apiserver/src/main/java/org/dependencytrack/persistence/jdbi/FindingDao.java
+++ b/apiserver/src/main/java/org/dependencytrack/persistence/jdbi/FindingDao.java
@@ -608,6 +608,9 @@ public interface FindingDao {
             String[] filters = filter.split(",");
             for (int i = 0, length = filters.length; i < length; i++) {
                 queryFilter.append(column).append(" = :").append(paramName).append(i);
+                if (paramName.equals("severity")) {
+                    queryFilter.append("::SEVERITY");
+                }
                 params.put(paramName + i, filters[i].toUpperCase());
                 if (filters[i].equals("NOT_SET") && (paramName.equals("analysisStatus") || paramName.equals("vendorResponse"))) {
                     queryFilter.append(" OR ").append(column).append(" IS NULL");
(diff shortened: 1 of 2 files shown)


apiserver/src/main/java/org/dependencytrack/persistence/jdbi/FindingDao.java
```diff
@@ -610,7 +607,7 @@ private void processArrayFilter(StringBuilder queryFilter, Map<String, Object> p
             queryFilter.append(" AND (");
             String[] filters = filter.split(",");
             for (int i = 0, length = filters.length; i < length; i++) {
-                queryFilter.append(column).append(" = :").append(paramName).append(i);
+                queryFilter.append(column).append("::TEXT = :").append(paramName).append(i);
```
nscuro: Casting all columns to `TEXT` could turn out to be a footgun later. At the moment this method is only called for enum columns that are not indexed, but if that changes later, casting the column could cause indexes to not be used.

There is already a `paramName` check further down. We should do something similar here, e.g.:

```java
if (paramName.equals("severity")) {
    queryFilter.append(column).append(" = :").append("%s::SEVERITY".formatted(paramName + i));
}
```

That way the casting happens on the parameter side.

nscuro (changes requested):

nscuro:
Thanks for the PR @ider-criteo! Please also make sure to [sign-off your commits](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--s).

apiserver/src/main/java/org/dependencytrack/persistence/jdbi/FindingDao.java
ider-criteo: Changed. Is it okay this way?

ider-criteo:
> Thanks for the PR @ider-criteo! Please also make sure to [sign-off your commits](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--s).

Sorry about that. Signed off now.

nscuro (approved):

PR #2055, opened 2026-04-28
Port : Add component scope support and CDX mapping

sahibamittal:
### Description

Add component scope.

### Addressed Issue

https://github.com/DependencyTrack/hyades/issues/2105
Frontend: https://github.com/DependencyTrack/hyades-frontend/pull/501
Ports https://github.com/DependencyTrack/dependency-track/pull/5224

### Checklist

- [ ] This PR is a substantial change (per the [ADR criteria]), and I have added an [ADR] under `docs/adr/`

[ADR]: ../docs/adr/


migration/src/main/resources/migration/changelog-v5.7.0.xml
```diff
@@ -1989,4 +1989,13 @@
...
         </sql>
     </changeSet>
+
+    <changeSet id="v5.7.0-82" author="sahibamittal">
+        <addColumn tableName="COMPONENT">
+            <column name="SCOPE" type="TEXT"/>
+        </addColumn>
+        <createIndex tableName="COMPONENT" indexName="COMPONENT_SCOPE_IDX">
+            <column name="SCOPE"/>
+        </createIndex>
```
nscuro: The index is unlikely to be useful given the low cardinality of the column. I'd say we omit if for now, we can always add it when it turns out to be needed.

However, we should add a check constraint that verifies the allowed scope values, similar to what we do for classifiers.

nscuro (changes requested):
We should also update the schemas for API v2 to include the new field, e.g. https://github.com/DependencyTrack/hyades-apiserver/blob/ac3ec3bdf324a98d01d849ef6f7bad8537dca375/api/src/main/openapi/resources/components/schemas/list-components-response-item.yaml

api/src/main/openapi/resources/components/schemas/list-components-response-item.yaml
```diff
@@ -28,6 +28,9 @@ properties:
   classifier:
     type: string
     maxLength: 255
+  scope:
+    type: string
+    maxLength: 255
   hashes:
```
nscuro, line 33: Scope should be an enum.

migration/src/main/resources/migration/changelog-v5.7.0.xml
```diff
@@ -1989,4 +1989,14 @@
              WHERE "NAME" = 'VIEW_BADGES';
         </sql>
     </changeSet>
+
+    <changeSet id="v5.7.0-82" author="sahibamittal">
```
nscuro, line 1993: A changeset for `-82` has since been merged, please increment to `-83`.

nscuro (changes requested):
Two tiny changes, otherwise good to go!

nscuro (approved):

PR #2138, opened 2026-05-11
Rename container image and add tag policy

### LINKED ISSUE #6103: Rename v5 apiserver image and add v5 tag policy
This issue is a single PR on `DependencyTrack/hyades-apiserver:main`. It ships the image rename, the v5 tag policy, and
the major-version derivation as one atomic change, with a short preflight that gates the merge. These pieces cannot be
split across PRs without leaving a window where `:snapshot` collides with v4 nightly or `:5` and `:5.0` are not emitted at
all.

### 1. Run the preflight that gates the merge

Two checks must come back green **before** the PR merges, because the moment it merges the snapshot publish destination
flips.

First, audit the Docker Hub credentials. Confirm that `DOCKER_HUB_TOKEN` and `DOCKER_HUB_USERNAME` in the
`hyades-apiserver` repo secrets have **push** rights on `dependencytrack/apiserver`. Today the token pushes to
`dependencytrack/hyades-apiserver`. After this PR merges, the mirror workflow starts pushing to the new destination on
every snapshot build. If the token is repo-scoped, mirroring silently fails. Re-scope or re-issue the token first.

Second, pre-create the GHCR package. The first push to `ghcr.io/dependencytrack/apiserver` auto-creates the package as
**private**. Public users pulling `:5-snapshot` would get auth errors. Pre-create the package, set its visibility to
**public**, and link it to `hyades-apiserver` for now. Re-link it to `dependency-track` when the apiserver cutover changes
the source repo.

### 2. Add caller-side major-version derivation in `ci-build.yaml`

Add a `read-version` job to `ci-build.yaml`. Mirror the shape of the existing `read-version` job in `ci-publish.yaml`
(lines 28-49), but use Maven-native tooling instead of `yq`:

```
VERSION=$(mvn -q -Dexec.skip=true help:evaluate -Dexpression=project.version -DforceStdout)
MAJOR="${VERSION%%.*}"
```

Output `major` from the job. Pass `app-version: "${MAJOR}-snapshot"` to `call-build` instead of the literal `"snapshot"`.
For v5 main this resolves to `5-snapshot`.

In the same PR, switch `ci-publish.yaml`'s existing `read-version` job from `yq` to `mvn help:evaluate`. One Maven-native
pattern is easier to read than two parallel ones.

### 3. Rename the image in `_meta-build.yaml`

Change the hardcoded `IMAGE_NAME="ghcr.io/${GITHUB_REPOSITORY_OWNER,,}/hyades-apiserver"` to use `apiserver`. That is the
only file that needs the rename. The Docker Hub mirror destination in `mirror-container-image.yml` is auto-derived from
the GHCR package via `${PACKAGE_NAMESPACE}/${PACKAGE_NAME}`, so it follows automatically.

### 4. Add the v5 tag policy in `_meta-build.yaml`

Parse the major from `APP_VERSION`. When the major equals `5`, the tag set is:

- For release builds where `APP_VERSION` is a clean semver like `5.0.0`: emit `:5.0.0`, `:5.0`, and `:5`. Parse the major
  and minor with Bash parameter expansion or `cut`. Do **not** emit `:latest`.
- For pre-release builds where `APP_VERSION` is something like `5.0.0-rc.1`: emit `:5.0.0-rc.1` only. Do not emit `:5`,
  `:5.0`, or `:latest`.
- For snapshot builds where `APP_VERSION` is `${MAJOR}-snapshot`: emit `:${APP_VERSION}` literally. Bare `:snapshot`
  suppression is automatic, because after the caller-side change `APP_VERSION` is never the literal string `"snapshot"`.
  Do not emit `:main`.
- For feature-branch builds: leave behavior unchanged. The existing `:feature-*` tag is fine.

When the major is not 5, behavior is unchanged. v4's existing pipeline on `dependency-track:4.14.x` keeps emitting
`:latest` and (with the same caller change carried over later) `:4-snapshot`.

> [!NOTE]
> Feature branches push `:feature-${name}`. Today, v4 (`dependencytrack/apiserver`) and v5
(`dependencytrack/hyades-apiserver`) publish to different image repos and cannot collide. After cutover they both publish
to `dependencytrack/apiserver`, and a feature branch with the same name on both `4.14.x` and v5 `main` would then collide.

### 5. Verify before merge

Run `workflow_dispatch` dry-runs against synthetic inputs with
`publish-container=false`. The release pipeline rehearsal issue runs these tests too. The verification on this PR exists
to gate the merge.

Confirm:

- A synthetic `5.0.0` produces `:5.0.0`, `:5.0`, `:5`. No `:latest`.
- A synthetic `5.0.0-rc.1` produces `:5.0.0-rc.1` only.
- A synthetic v5 `main` snapshot produces `:5-snapshot` only. No bare `:snapshot`.

After merge, the next push to `hyades-apiserver:main` does a real `:5-snapshot` push. Verify the tag resolves on both
`ghcr.io/dependencytrack/apiserver` and `dockerhub/dependencytrack/apiserver`, and that both are public.

Once this issue is merged, the release pipeline rehearsal can
start, and the apiserver cutover can be scheduled.

nscuro:
### Description

Implements https://github.com/DependencyTrack/dependency-track/issues/6103 by:

1. Renaming the container image from `hyades-apiserver` to `apiserver`.
2. Publishing snapshot versions as `<major>-snapshot` instead of `snapshot` to avoid conflicts with v4.
3. For release builds, publishing `:<major>.<minor>.<patch>`, `:<major>.<minor>`, and `:<major>` tags. `latest` is no longer published.

### Addressed Issue

Closes https://github.com/DependencyTrack/dependency-track/issues/6103

### Additional Details

N/A

### Checklist

- ~This PR is a substantial change (per the [ADR criteria]), and I have added an [ADR] under `docs/adr/`~

[ADR]: ../docs/adr/
[documentation]: https://github.com/DependencyTrack/docs
[migration changelog]: ../DEVELOPING.md#database-migrations

### DIFF
--- a/.github/workflows/_meta-build.yaml
+++ b/.github/workflows/_meta-build.yaml
@@ -19,8 +19,7 @@ on:
     inputs:
       app-version:
         type: string
-        required: false
-        default: "snapshot"
+        required: true
         description: "the version that should be set/used as tag for the container image"
       publish-container:
         type: boolean
@@ -123,12 +122,10 @@ jobs:
       matrix:
         include:
           - name: apiserver
-            image: hyades-apiserver
             context: apiserver
             artifact: assembled-dist
             prepare-cmd: tar xzf apiserver/target/dependency-track-apiserver-dist.tar.gz -C apiserver/target
           - name: v4-migrator
-            image: v4-migrator
             context: support/v4-migrator
             artifact: v4-migrator-jar
             prepare-cmd: ":" # no-op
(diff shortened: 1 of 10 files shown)


nscuro:
Docker Hub credentials have been updated to a token that has push permissions for both `apiserver` and `hyades-apiserver` repositories.

nscuro:
The `ghcr.io/dependencytrack/apiserver` GitHub package cannot be pre-created manually it seems. First push after merge will auto-create it. Just need to ensure to flick it to public afterwards.

PR #2197, opened 2026-05-25
Fix pending partition detach issue

sahibamittal:
### Description

When `DETACH PARTITION ... CONCURRENTLY` is interrupted between two internal phases, the partition remains in 'pending detach' state. This may cause in metrics maintenance task failure.

This change will check before attempting the detach, if the partition is already pending, and use `DETACH PARTITION ... FINALIZE` to complete it instead of re-issuing CONCURRENTLY.

Adds a test to simulate a partition in pending-detach state and test it.

### Addressed Issue

NA

### Checklist

- [ ] This PR is a substantial change (per the [ADR criteria]), and I have added an [ADR] under `docs/adr/`

[ADR]: ../docs/adr/
[documentation]: https://github.com/DependencyTrack/docs
[migration changelog]: ../DEVELOPING.md#database-migrations


apiserver/src/main/java/org/dependencytrack/persistence/jdbi/MetricsDao.java
```diff
@@ -605,14 +605,32 @@ default int dropPartitions(final String parentTable, final List<String> partitio
         int deletedCount = 0;
         for (final String partition : partitions) {
             requireValidTableIdentifier(partition);
-            getHandle().execute("ALTER TABLE %s DETACH PARTITION %s CONCURRENTLY".formatted(parentTable, partition));
+            if (isPartitionDetachPending(parentTable, partition)) {
+                getHandle().execute("ALTER TABLE %s DETACH PARTITION %s FINALIZE".formatted(parentTable, partition));
```
nscuro: IIUC, this can now block for potentially unlimited time. It would be good to limit the wait time so the connection isn't blocked for too long.

To my understanding, executing `SET lock_timeout = '5s'` (for example) *before* the actual statement should do the trick, see https://postgresqlco.nf/doc/en/param/lock_timeout/

I believe `SET` only works when in a transaction, so would need to make sure a trx is active when executing this.

nscuro (approved):

